PR #3, opened 2015-12-07
A couple of fixes

jensnielsen:
Hi

I don't know how to create individual pull requests for each commit, so here are three commits that are more or less harmless...

First one just fixes some compile problems I had in my project. It's pretty straightforward, even though you might want to do some of it differently

Second one is for https://sourceforge.net/p/canopennode/tickets/3/ but it's not really needed anymore since it's already initialized to 0 since some time back. I thought it was nice to have anyway.

Third one is about https://sourceforge.net/p/canopennode/discussion/387151/thread/de4ce352/ (which is still awaiting moderation?). It's a bit trickier but if you compare without whitespace it should be a lot easier to see what's going on. (Handle all buffers in rx interrupt instead of just the one indicated by ICODE)

Best regards
Jens

### DIFF
--- a/CANopen.c
+++ b/CANopen.c
@@ -173,7 +173,7 @@ CO_ReturnError_t CO_init(
 #ifdef CO_USE_GLOBALS
     CO = &COO;
 
-    CO->CANmodule[0]                    = &COO_CANmodule[0];
+    CO->CANmodule[0]                    = &COO_CANmodule;
     CO_CANmodule_rxArray0               = &COO_CANmodule_rxArray0[0];
     CO_CANmodule_txArray0               = &COO_CANmodule_txArray0[0];
     CO->SDO                             = &COO_SDO;
--- a/stack/CO_NMT_Heartbeat.c
+++ b/stack/CO_NMT_Heartbeat.c
@@ -192,6 +192,7 @@ CO_NMT_reset_cmd_t CO_NMT_process(
 
         if(NMT->operatingState == CO_NMT_INITIALIZING){
             if(HBtime > NMT->firstHBTime) NMT->HBproducerTimer = HBtime - NMT->firstHBTime;
+            else NMT->HBproducerTimer = 0;
 
             if((NMTstartup & 0x04) == 0) NMT->operatingState = CO_NMT_OPERATIONAL;
             else                         NMT->operatingState = CO_NMT_PRE_OPERATIONAL;
--- a/stack/PIC24H_dsPIC33F/CO_driver.c
+++ b/stack/PIC24H_dsPIC33F/CO_driver.c
@@ -615,74 +615,103 @@ void CO_CANverifyErrors(CO_CANmodule_t *CANmodule){
     }
 }
 
+static void CO_CANHandleRx(CO_CANmodule_t *CANmodule, uint8_t bufferNumber)
+{
+    EDS_PTR CO_CANrxMsg_t *rcvMsg;/* pointer to received message in CAN module */
+    uint16_t index;             /* index of received message */
+    uint16_t rcvMsgIdent;       /* identifier of the received message */
+    CO_CANrx_t *buffer = NULL;  /* receive message buffer from CO_CANmodule_t object. */
+    bool_t msgMatched = false;
+    volatile uint16_t C_CTRL1old;
 
+    rcvMsg = &CANmodule->CANmsgBuff[bufferNumber];
+    rcvMsgIdent = rcvMsg->ident;
+    if(CANmodule->useCANrxFilters){
+        /* CAN module filters are used. Message with known 11-bit identifier has */
+        /* been received */
+        index = rcvMsg->FILHIT;
+        if(index < CANmodule->rxSize){
+            buffer = &CANmodule->rxArray[index];
+            /* verify also RTR */
+            if(((rcvMsgIdent ^ buffer->ident) & buffer->mask) == 0U){
+                msgMatched = true;
+            }
+        }
+    }
+    else{
+        /* CAN module filters are not used, message with any standard 11-bit identifier */
+        /* has been received. Search rxArray form CANmodule for the same CAN-ID. */
+        buffer = &CANmodule->rxArray[0];
+        for(index = CANmodule->rxSize; index > 0U; index--){
+            if(((rcvMsgIdent ^ buffer->ident) & buffer->mask) == 0U){
+                msgMatched = true;
+                break;
+            }
+            buffer++;
+        }
+    }
+
+    /* Call specific function, which will process the message */
+    if(msgMatched && (buffer != NULL) && (buffer->pFunct != NULL)){
+#ifdef __HAS_EDS__
+        CO_CANrxMsg_t _rcvMsg = *rcvMsg;
+        buffer->pFunct(buffer->object, &_rcvMsg);
+#else
+        buffer->pFunct(buffer->object, rcvMsg);
+#endif
+    }
+
+    /* Clear RXFUL flag */
+    CO_DISABLE_INTERRUPTS(); //TODO
+    C_CTRL1old = CAN_REG(CANmodule->CANbaseAddress, C_CTRL1);
+    CAN_REG(CANmodule->CANbaseAddress, C_CTRL1) = C_CTRL1old & 0xFFFE;     /* WIN = 0 - use buffer registers */
+    if(bufferNumber < 16) CAN_REG(CANmodule->CANbaseAddress, C_RXFUL1) &= ~(1 << bufferNumber);
+    else         CAN_REG(CANmodule->CANbaseAddress, C_RXFUL2) &= ~(1 << (bufferNumber & 0xF));
+    CAN_REG(CANmodule->CANbaseAddress, C_CTRL1) = C_CTRL1old;
+    CO_ENABLE_INTERRUPTS();
+}
 /******************************************************************************/
 void CO_CANinterrupt(CO_CANmodule_t *CANmodule){
-    uint16_t ICODE;
-    ICODE = CAN_REG(CANmodule->CANbaseAddress, C_VEC) & 0x1F;
 
-    /* receive interrupt (New CAN messagge is available in RX FIFO buffer) */
-    if(ICODE > 0){
-        EDS_PTR CO_CANrxMsg_t *rcvMsg;/* pointer to received message in CAN module */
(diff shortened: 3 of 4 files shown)


CANopenNode:
Hi Jens,

I didn't successfully compile the new code for dsPIC33 yet, so first commit is necessary.

Second issue - it seems there was a non-optimal behaviour. I suggest following solution:

        if(NMT->operatingState == CO_NMT_INITIALIZING){
            if(HBtime > NMT->firstHBTime) {
                NMT->HBproducerTimer = HBtime;
            } else {
                NMT->HBproducerTimer = NMT->firstHBTime;
            }

Third issue: I was never actively using dsPIC33 or PIC24 and documentation is not very clear. Solution is a little dirty, because it is high frequency interrupt and there are two for loops 16x. Before update I will check documentation a little.

(The thread on sourceforge was waiting for moderation for some reason. I didn't got a notice.)

Regards,
Janez

jensnielsen:
Hi!

First commit: Thanks!

Second: Hmm this is that really what you want? NMT->HBproducerTimer is counting upwards and when it exceeds HBtime a new heartbeat will be sent, so basically you want it to be `HBtime - min(HBtime, NMT->firstHBTime)`, right? I.e. zero if HBtime is less than NMT->firstHBTime

Third: Please do! I couldn't find anything useful in either documentation or examples from Microchip, but maybe I'm looking in the wrong places... Yes I agree solution is dirty, I'm also having some problems with it taking too long so I'd be happy to avoid the loop... One obvious improvement is to only loop over `CANmodule->CANmsgBuffSize`, for example. I was also thinking of doing something smarter with filters (I have more than 16 rx buffers) but I guess that would be a less generic solution and only suitable for my project.

Best regards
Jens

CANopenNode:
Hi,

Second: sorry, my mistake. (I missed the logic of original solution.) Your solution is just fine.

Third: I checked your code again. I have a simple proposal, how to avoid the two for loops in most cases. Just add the line `CO_CANHandleRx(CANmodule, ICODE);` in the beginning, before you read the `C_RXFUL1copy` and `C_RXFUL1copy`. Then run each for loop only, if ether of them is different than zero. This should reduce processor time of this high priority interrupt significantly.
If you add and test it, I will just make a merge.

jensnielsen:
Hi

Alright, sorry, I didn't see that last comment. I just pushed another variant where I only loop over CANmsgBuffSize, but doing ICODE first is a good idea though, I'll add that too.

After browsing the documentation some more I think we're supposed to use CiFIFO<FBP> and CiFIFO <FNRB> for looping, I'll look into that some more.

There's something fishy with using more than 8 buffers though, I get RXFUL flag set but there's nothing in the DMA buffers (all zeroes).. I don't think it's related to my change and all registers looks fine to me (DMABS is correct, FBP points to buffer > 8, etc) but I feel like I'm missing something obvious...

Regards,
Jens

CANopenNode:
Hi,

FIFO may be a good idea. I'm not sure, if it existed few years ago, when I was writing that code.
I'm using FIFO on PIC32 and it works perfectly.

jensnielsen:
Hm, my last post was cut by markup, it's supposed to say "`CiFIFO<FBP>` and `CiFIFO<FNRB>`"..

We are using FIFO already, what is unclear is just how to know which buffers to read. There's some information in reference manual section 21 "ECAN" chapter 21.7.3 "FIFO Buffer Operation" and 21.7.4 "FIFO Example", but it's not enough. I can only conclude that Microchip didn't think this through...

According to the reference manual we're supposed to read the messages in sequence, starting with FNRB. When a message is serviced we clear RXFUL and then FNRB is updated automatically to tell us next message. My approach was to do this in a loop until either RXFUL is not set for the buffer indicated by FNRB or we have checked all RX buffers. This works for the normal scenario but there's no way to recover from an overflow scenario, since FBP continues to increment. After an overflow, if one tries to recover, FNRB will point to an unused buffer while FBP writes to another and there's no way to move FNRB. So we are given three options: 1) an efficient interrupt that doesn't work after overflow 2) an inefficient interrupt that continues to work after overflow, or 3) use the efficient interrupt handler but reset controller at overflow.

I don't like option 3 (the only solution to a lost message is to lose more messages?) so I'm leaning towards option 2...

CANopenNode:
What is the possibility of overflow?
I think, if CAN interrupt has highest priority in the whole application, then microcontroller has enough time to handle all received messages, also 8 buffers should be enough. If there are large higher priority delays inside application, then there may be a problem. Slow interrupt may have more problems (even large buffer will fill), than small buffer.
Can you do a test under heavy bus traffic in 1Mbps and measure processor time spent for interrupt?
I made such test on 8-bit pic some time ago, without hardware filters, and it worked fine. It had only two rx buffers I think.

jensnielsen:
Yes I also think an overflow is unlikely to happen in a normal scenario, but it happens for me every time I break with the debugger. I don't know why, everything is checked under "Freeze peripherals" in MPLAB so I expect the CAN controller to freeze, but it's quite annoying if CAN communication breaks every time I'm debugging... And it's also a matter of professionalism that I want to make a correct implementation :)

However, if I just ignore all messages until FBP catches up with FNRB again it will either recover or trigger another overflow. The messages will be handled in the wrong order and it may take a few iterations but eventually it will recover. I suppose that's good enough, I will push a new proposal.

CANopenNode:
I also don't have good experience with debugger and CAN module. So I just don't trust it and I try not to use it if not really necessary. Instead I try to catch some strange behaviour with code in run time and store the result in a variable. Then variable can be accessed by debugger or by CANopen, if it is in object dictionary. Or even map the variable(s) in PDO and then examine CAN message log.

You can try to catch the maximum number of messages in FIFO buffer and then watch that variable in real or overloaded operating conditions.
If your application doesn't disable interrupts(CANrx) for too long time, fifo buffer should never have
more than two or three messages inside.

I suggest not to use 32 fifo buffers and use max 16. Interrupt will be a little simpler and faster, if it won't check additional C_RXFUL2 register. Stack itself should be efficient enough to handle this.

I have another proposal. (I didn't look into datasheet, so I'm not sure.) If we use the original code (without for loops) and add another check **after** clearing interrupt flag `CAN_REG(CANmodule->CANbaseAddress, C_INTF) &= 0xFFFD;`. Check the `C_RXFULx` flag and if different than zero, re-enable interrupt bit `C_INTF`. Will it trigger interrupt again and handle next message? Can you test it?
(My suggestion for testing without debugger: Make a source of CAN messages with interval 100ms for example. Then make a delay in your application, where you disable interrupts for 500ms. After re-enabling interrupts fifo should have 4-6 messages and interrupt should process all of them. If you have oscilloscope, you can use a pin, which you set high, when rx interrupt is executing. There you can see, if everything is OK and you can also measure how much time is spent by interrupt.)

jensnielsen:
Yes, we also have a display on this product where I can print debug info, it's useful especially for troubleshooting performance related issues like this, but sometimes it's so much easier to step with a debugger to figure out logical problems (and problems not related to CAN) so we have to be able to use that.

We're only using 8 buffers now, but this is an input parameter to CO_CANmodule_init() so the driver should handle values larger than 16. But maybe we can make it a global define and exclude RXFUL2 in compile time if it's 16 or lower? I.e. move CO_CANmsgBuffSize from CANopen.c to some project header file? I have worked with other communication stacks where there's a mandatory project header file where you can configure non-default options, I think this would be a good thing here too. CO_CANmsgBuffSize, CO_USE_GLOBALS, and DMA channels are some examples of defines that could go in there.

You mean the first original code? One of my first tests was to not clear C_INTF if there are more messages pending (which would be the same thing, right?) but that doesn't work since ICODE will always point to the latest received message. Using that approach but with FNRB instead of ICODE will be pretty much the same as I have now except that the interrupt will be called for each message instead of handling them in a loop, so it will be more overhead. Note that I added a break in the loop now when RXFUL is 0 so the loop will only handle pending messages and stop looping on the first empty buffer.

CANopenNode:
This is a critical bugfix, so I will make some tests on my Explorer16 board before updating the code. Hopefully, this week.

jensnielsen:
Hi
Thanks! Unfortunately my proposed solution didn't work well in a high load scenario... Here's another solution (avoid reading past FBP)
Also fixed an issue with RXFUL2, though that part is still untested...

CANopenNode:
Hi Jens,

I made a tests on my Explorer16 board. ICODE was a bad solution, FIFO didn't work, which was serious bug causing RX overflow. It didn't show many times, because interrupt was quite fast.

I think your solution with FNRB is best choice. I merged all your code and made some changes. (Actually I didn't merge directly, so your fork quite differs).

Please test my version. FIFO now works. In case of overflow, CANrx recovers successfully, also when using debugger (same as in your code).

I change one thing, which increases response, if there is large FIFO buffer. First I limited size of FIFO to 16 messages, so RXFUL2 is avoided. Then I used FNRB directly without for loop. (for loop is only for fallback (debugger break).)

Interrupt time for complete processing of one CANrx (PDO) message is now 12us (10us was with ICODE). I didn't test for high load scenario, but I think it should work.

Janez

CANopenNode:
I tested also the FIFO in CANrx for PIC32. There is FIFO for 32 messages. It uses ICODE register, to determine, which buffer contains a new message. (Same solution, as is was in dsPIC33, before bugfix.) 

However, FIFO and interrupt triggering works perfectly there.

jensnielsen:
Hi

Sorry for extremely slow response, I've been a bit distracted by other things and as always panic level increases as deadline gets closer...

I have however been using your version of the driver for a while without seeing any problems so it seems to be fine. I'm not sure I follow what the improvement with the loop (in the "This should not happen" if-statement) was over my version though. I wasn't looping over all buffers either, in the normal case only until I hit the first empty buffer. I think your version won't recover completely from the overflow situation, if you do get the FNRB/FBP out of sync issue I think you will always end up in the "This should not happen" case and have to loop to find the correct buffer for all incoming messages, while my version should recover completely but is worse at handling the situation while recovering (it may throw away some extra messages and handle the buffers in the wrong order). It's just a matter of taste though so I won't object to your choice.

I agree avoiding to use RXFUL2 sounds like a good choice

jensnielsen:
... Or maybe I'm wrong, in my mind FNRB would continue to increment in the "This should not happen" case and stay out of sync with FBP, but it shouldn't so your solution should be better than mine. I will add some counters and test a bit more thoroughly. Sorry for the confusion...

PR #4, opened 2015-12-07
Allow multiple SDO objects

jensnielsen:
These are the changes I mentioned for multiple SDO servers. (At least as I understand how it should work)

Unless I'm mistaken the first server must be the default one (COB 0x580/0x600 + node id) but the others can be whatever you want

I also added a parameter "offset" to ODF_arg_t to allow multiple simultaneous reads of the same Domain OD object, this parameter should replace the static variable you had in ODF callback function. The ODF callback still has to take care of multiple simultaneous writes though, that might cause headache...

Best regards
Jens


CANopenNode:
Thanks for this.

PR #8, opened 2016-03-03
CAN RX interrupt

jensnielsen:
Hi again

Yes you were right, I'm reverting to your version of the interrupt, except for one thing. You don't need my if (FNRB == FBP) break at the end of the loop, it will recover anyway, so here's a commit to remove this. This should also reduce the number of dropped messages during recovery.

I have verified and seen that it will just enter the "This should not happen" case a couple of times until FBP catches up with FNRB and then it will be back to normal again.

Thanks for all the help!


jensnielsen:
Oops sorry for my lack of github skills, the diff is correct but there's only supposed to be one commit here (the last one)

BR /Jens

CANopenNode:
Hi Jens,

I updated the code with your last commit. Thank you.

Issue #14, opened 2016-10-14
Can't write to subindex zero of register 0x1003 Pre-defined error field


Drllap:
It is my understanding that this should be possible, and doing it should reset the error counter.

I also found [this](https://github.com/CANopenNode/CANopenNode/blob/5df11b9d749b3802267f4c6d50259eb427c6139a/stack/CO_Emergency.c#L81) piece of code that suggests this has been thought of at some point.

If I try to do this write, the function above never gets executed, instead CO_SDO_AB_READONLY error is thrown [here](https://github.com/CANopenNode/CANopenNode/blob/5df11b9d749b3802267f4c6d50259eb427c6139a/stack/CO_SDO.c#L816)

This happens even though OD entry for this register in initialised like [this](https://github.com/CANopenNode/CANopenNode/blob/5df11b9d749b3802267f4c6d50259eb427c6139a/example/CO_OD.c#L306) with attribute of 0x8E and never changes. (Note that CO_ODA_READABLE == 0x04)

What happens is that when the object for this registered is fetched from the OD the attribute member is populated with [this](https://github.com/CANopenNode/CANopenNode/blob/5df11b9d749b3802267f4c6d50259eb427c6139a/stack/CO_SDO.c#L485) function that makes sure that subindex zero is never written to.

CANopenNode:
I'm currently short with my time. I have to update and test index 0x1003 more precisely. I plan to do this in ten days.

CANopenNode:
There was a problem with suggested fix, as described in comment to #15.

However, it is fixed now.
Thanks

PR #15, opened 2016-10-21
Make it possible to write to 0x1003 - Pre-Defined Error Field

Drllap:
My fix for #14. A bit of a hack I know, but you can have it if you want it


CANopenNode:
There is a problem with this pull request, because 1003,00 gets write attribute, but associated variable should not be written.

However, I solved the problem the similar way, but inside the CO_SDO_writeOD() function. See https://github.com/CANopenNode/CANopenNode/commit/83b6a7abb4e5df72f915e21c6327d6afa3168c10

PR #16, opened 2016-11-23
Allow multiple SDO client objects

WillyKaze:
These changes allow multiple SDO client objects to be declared in CO_OD.h

This is hugely based on #4.

I'd like your feedback, I'm quite new to the CANopen world and maybe I missed a way to implement this with one SDO client and multiple SDO objects.

Best regards
Ben

### DIFF
--- a/CANopen.c
+++ b/CANopen.c
@@ -137,7 +137,7 @@
     static CO_HBconsumer_t      COO_HBcons;
     static CO_HBconsNode_t      COO_HBcons_monitoredNodes[CO_NO_HB_CONS];
 #if CO_NO_SDO_CLIENT != 0
-    static CO_SDOclient_t       COO_SDOclient[CO_NO_SDO_CLIENT];
+    static CO_SDOclient_t       COO_SDOclient;
 #endif
 #if CO_NO_TRACE > 0
     static CO_trace_t           COO_trace[CO_NO_TRACE];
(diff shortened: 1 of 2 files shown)


jensnielsen:
Hi

I haven't worked with CANopen for a while so I might be missing something and I don't know if you need to do something different, this just caught my attention since you referenced my #4 

One thing I noticed is when calling CO_SDOclient_init() you are passing CO->SDO[i] but I think this should still be CO->SDO[0] shouldn't it? This parameter would be the "local server" the client will communicate with in case the SDO targets the local node. SDO[0] is guaranteed to exist but by passing in SDO[i] you assume you have as many servers as you have clients and each client will have its own server. Or is that intended?

Just out of curiosity, what's your use case? I think I may have solved the same problem by simply having a queue with pending requests and a task interfacing my single SDO client

Best regards
Jens

WillyKaze:
Hi,

You are correct, we have to use the default SDO[0].
I made progress yesterday and I actually reverted some of my changes:

The goal was to create multiple SDOclient entries in the OD so I started with multiple SDOclient object but the SDOclient cli functions already allow to specify different COB-ID to send/receive.

Today, I use multiple SDOclient COB-ID but only one SDOclient. I need OD entries because the COB-ID for SDO communication are configured by the master node at startup.

Best regards
Ben

CANopenNode:
Hi,

For SDO client with multiple nodes you don't need anything extra. See for example here:
https://github.com/CANopenNode/CANopenSocket/blob/master/canopend/src/CO_master.c

Your driver must allow reconfiguration of CAN receive filters in run time.

WillyKaze:
Hi,

I reverted my changes for multiple SDOclient. Right now the only modification is to allow CO_NO_SDO_CLIENT to be over 1.
This is allow to creating multiple entries in OD_SDOClientParameter array.

Tell me if it is still misleading or useless.

Best regards,
Ben

CANopenNode:
Hi,

CANopen.[ch] files are meant to be universal for usual device configurations. For a special configuration they can be customized, but I don't like for them to be to complicated.

I have plans to renew this part of CANopenNode, together with object dictionary in near future.
Currently I won't merge this pull request.

Best regards, 
Janez

PR #53, opened 2017-08-30
Initial commit

antony-sm:


martinwag:
Thanks, this is great! But I think this is a project in it's own and not an addition to CANopenNode.
- complete rewrite
- API incompatible
- C++ instead of C
- needs specific OS (freeRTOS)
which means it would actually have fulfilled my own requirements if it had been around then :-)

PR #59, opened 2017-09-20
add S32 SDK support

bubi-soft:


CANopenNode:
I was inactive with CANopenNode for two years, so I missed this commit, see #105.

However, this driver is added to deviceSupport list for CANopenNode. See "doc/deviceSupport.md" - https://github.com/CANopenNode/CANopenNode/blob/split-driver/doc/deviceSupport.md (currently this is split-driver branch).

Please note also, license of the project changed to more permissive Apache2.

Thank you for your contribution.

PR #67, opened 2017-12-06
Emergency Message RX Callback

wilkinsw:
A feature which I've found lacking was the inability to take action when another node on the network transmits an emergency message.  These commits expand the emergency message callback to function similarly to the NMT callback.  Instead of this callback being called only on transmission of an emergency message it is now called on reception of one as well.

For example, we wanted our master (in this case canopend) to report any emergency messages on the bus by passing them through the command socket.  

First we define a handler:
```
/* Handle CANopen emergency messages. */
void EM_report(const uint16_t ident,
               const uint16_t errorCode,
               const uint8_t errorRegister,
               const uint8_t errorBit,
               const uint32_t infoCode) {

    char buf[40];
    int len;

    len = sprintf(buf,
                  "EM: %04x %04x %02x %02x %08x\r\n",
                  ident,
                  errorCode,
                  errorRegister,
                  errorBit,
                  infoCode);

    CO_command_write(buf, len);
    taskMain_cbSignal();
}
```

Next we register it during setup:
`CO_EM_initCallback(CO->em, EM_report);`

Now whenever one of our nodes has a fault it is sent to the socket so action can be taken.

### DIFF
--- a/CANopen.c
+++ b/CANopen.c
@@ -106,12 +106,13 @@
 
     #define CO_RXCAN_NMT       0                                      /*  index for NMT message */
     #define CO_RXCAN_SYNC      1                                      /*  index for SYNC message */
-    #define CO_RXCAN_RPDO     (CO_RXCAN_SYNC+CO_NO_SYNC)              /*  start index for RPDO messages */
+    #define CO_RXCAN_EMERG    (CO_RXCAN_SYNC+CO_NO_SYNC)              /*  index for Emergency message */
+    #define CO_RXCAN_RPDO     (CO_RXCAN_EMERG+CO_NO_EMERGENCY)        /*  start index for RPDO messages */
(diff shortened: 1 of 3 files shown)


flcapurso:
Any reason why this is not yet merged? I ran into the same issue and am currently using this code to handle incoming emergency messages.

wilkinsw:
Unfortunately I don't think anyone has heard from the owner since 2017, so the project is mostly abandoned.

wilkinsw:
Woops, didn't mean to close it.  Might as well keep it open just in case!

wilkinsw:
@flcapurso good news!  Looks like CANopenNode might be back in development soon!

see #105

martinwag:
I've changed that into two callbacks for compatibility of current software. please have a look as I don't use this.

PR #79, opened 2018-04-12
Fix shadowed variables warnings

odesenfans:
Fix warnings from -Wshadow.


odesenfans:
I started fixing compilation warnings. This is the first patch out of six.

CANopenNode:
I think, it is closed with #144

PR #106, opened 2019-09-23
Use inttypes.h for 32 bit signed/unsigned format specifiers.

henrikbrixandersen:
This fixes commit 533801574055c5dacf60fb6b18930c288336dce1 on 32-bit ARM with GCC 8.3.0 (among others).

### DIFF
--- a/stack/CO_trace.c
+++ b/stack/CO_trace.c
@@ -45,7 +45,7 @@
 
 #include "CO_trace.h"
 #include <stdio.h>
-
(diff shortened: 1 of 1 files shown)


martinwag:
This introduces need for C99 compiler. I have no idea if we already need one?

Can someone please give me a second opinion on that?

JuPrgn:
Hi Martin , snprintf is part of C99 standard.

PR #107, opened 2019-10-07
stack auto enters operational mode after init

geoffrey-vl:
I've noticed that when using the stack on slave nodes, the stack goes automatically into operational mode after boot. This happens when we setup the OD **1F80** record (which will contain the CANOpen state during boot) with value 0x00 (= CO_NMT_INITIALIZING).

The code that makes it automatically go into operational mode is in CO_NMT_Heartbeat.c; function CO_NMT_Process().
Currently we can adjust that behavior by setting the default value for the OD **1F80** record to 0x04 (= CO_NMT_STOPPED). 

However, that is not what CANOpen standard defines. From initialization mode the nodes are only allowed to enter pre-op mode. Only the master is allowed to set a node in operational mode:

[image CANOpen state machine](https://www.can-cia.org/fileadmin/resources/images/nmt-slave.gif)

I'd think the code would need to be as suggested:

### DIFF
--- a/stack/CO_NMT_Heartbeat.c
+++ b/stack/CO_NMT_Heartbeat.c
@@ -235,7 +235,18 @@ CO_NMT_reset_cmd_t CO_NMT_process(
             if(HBtime > NMT->firstHBTime) NMT->HBproducerTimer = HBtime - NMT->firstHBTime;
             else                          NMT->HBproducerTimer = 0;
 
-            NMT->operatingState = CO_NMT_PRE_OPERATIONAL;
+            
+            if((NMTstartup & 0x01) == 1) // 1F80 bit 0 indicates if node is NMT manager 
+            {
+                //NMT manager may enter operational mode automatically if bit 3 is set true
+                if((NMTstartup & 0x04) == 0) 
+                    NMT->operatingState = CO_NMT_OPERATIONAL;
+                else
+                    NMT->operatingState = CO_NMT_PRE_OPERATIONAL;
+            } else {
+                //slave node are only allowed to enter preop after boot
+                NMT->operatingState = CO_NMT_PRE_OPERATIONAL;
+            }
         }
     }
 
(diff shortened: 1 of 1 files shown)


robincornelius:
Hi  Geoffery,

According to the DSP-302 i have (v3.2.1) there are possible boot flowcharts
that show autostart of slave notes, section 4. (page 3) shows the use of
object 0x1f80, bit 2 to auto start a slave node that is not the NMT master.
So checking bit 2 of 0x1f80 and autostarting a slave is the correct thing
to do. I'm not sure if i've mis-understood the patch or the boot flow chart
;-)

Regards

Robin

geoffrey-vl:
Hi Robin,
thanks for referring to the docs, I hadn't checked that but was basing my knowledge from a course I was enrolled into, and google :)

Section 4 in that document states:

> When the CANopen Manager starts after Power-On, it will perform the state machine according to
DS-301. Before switching the state from Pre-Operational to Operational it has the task of booting all
assigned slaves. The overall process is shown in the following two flow charts.

The flowchart that is shown in sector 4 shows the path for CANopen manager. From what I understand, first thing the flowchart does, when booting / comms reset, is checking object 1F80h so to know that the node is configured as MNT master: if not node should enter slave mode and does not run any other parts of the flowchart.

Note that the NMT boot flowchart continues with Figure 2 on page 5. In that flowchart you'll notice that the NMT will only enter operational mode after all slaves have booted. If bit 2 is set, the NMT manager may enter operational state automatically after all slaves have booted. Otherwise it is up to the application to decide that. So I guess that's what the code reflects, but it is only applicable for master node, not slaves.

On sector 5.1, index 1F80h has following interpretation:

> Bit 0
= 0 Device is NOT the NMT Master. All other bits have to be ignored.
The objects of the Network List have to be ignored.
= 1 Device is the NMT Master.

So I guess my patch Is not correct either and needs to check that bit. I'll rework it.

martinwag:
I have current DSP 302-2, you need to be CiA Member to get it.

Self Start is described in 6.2 Object 1F80h, Table 16 — Exceptions for NMT start-up capable devices 

00000000 00000000 00000000 00001000b - self start

So current behaviour is OK. If any, you could read this as NMT master should ignore that bit. But NMT master can set its own state anyway...

geoffrey-vl:
So it seems the behavior has changed over time. Its a pitty that the DSP is being kept for Cia members only. I can't verify so I'll close the PR and use 0x04L as value for object 1F80h as that will give the same effect.

PR #110, opened 2019-10-09
feat: Add TimeStamp feature

JuPrgn:


martinwag:
Is this TS producer / consumer / both?

From what I can see this is only consumer, so to not have confusion when adding producer, should we change that to something like with heartbeat "CO_HBconsumer_t" -> CO_TSconsumer_t, "CO_RXCAN_CONS_HB" -> "CO_RXCAN_CONS_TS" and so on?

JuPrgn:
Yes this is only a consumer I will update this PR.

wilkinsw:
From what what I've been able to determine the Timestamp protocol consists of mapping 0x1013 to a TPDO on your producer and an RPDO on your consumers.

Could you provide an example of how your module would be initialized and used?

For example, where is CO_CAN_ID_TIME_STAMP defined?  You also say "For more information and description of parameters see file CO_driver.h."  But I don't see any changes to the CO_driver.h template.

JuPrgn:
From CiA 301 TIME can be consumer and or producer this is configured with object 0x1012. I don't see any usefull case to be a producer and consumer at the same time so I would prefer to choose while initializing this module and keep a single file but I am not sure about the best way to do this :

Would you prefer to keep a single module and I may add producer functions or separate this in two as you suggested ? 

@wilkinsw sorry this declaration is missing on the file I uploaded. Ok I will provide an example. For the comment this was already on the file I started from I think it was a copy of sync or heartbeat and I didn't modified this but there are no changes on driver part I am only adding an object to the dictionary. I made a PR for libedssharp to add TIME specific part but I will also update this OD example.

martinwag:
I've separated LSS in common, master and slave. In my opinion this was a good choise, even if it doesn't make sense to enable both.
Time protocol seems to be much simpler, so I think, do what you see is the best solution.

JuPrgn:
Added TIME producer, OD example and refactored many things. 

I did not tested this for the moment as I do not have any hardware for 2 weeks. I would appreciate if someone can give it a try. 

Please let me know if you need more informations about this.

You should set OD_COB_ID_TIME to : 
- 0x80000100L to be a TIME consumer
- 0x40000100L to be a TIME producer

JuPrgn:
I tested both producer and consumer and this is working as expected. 

**TIME CONSUMER mode**

CO_TIME_init() configuration :
- COB_ID_TIME : 0x80000100L -> TIME consumer with TIME_COB_ID = 0x100
- TIMECyclePeriod : 
     - 0 -> no EMCY will be transmitted in case of TIME timeout
     - X -> an EMCY will be transmitted in case of TIME timeout (X * 1.5) ms

Latest time value is stored in `CO->TIME->Time` variable.
 
**TIME PRODUCER mode**

CO_TIME_init() configuration :
- COB_ID_TIME : 0x40000100L -> TIME producer with TIME_COB_ID = 0x100
- TIMECyclePeriod : Time transmit period in ms

Write time value in `CO->TIME->Time` variable, this will be sent at TIMECyclePeriod.

martinwag:
Is this also available in the OD editor (enable/disable, producer/consumer, time of day struct)?

JuPrgn:
Hi Martin

This will be available soon in libedssharp I am waiting for Robin to merge the latest changes.

At the moment enable/disable and producer/consumer are selected by assigning a non-zero value to 0x1012 COB_ID_TIME.

blankm:
There's a little mistake I guess, the time object is never free'd.
In CO_delete all get's released but not the time object, why?

JuPrgn:
Thank you for pointed this out I will correct it !

blankm:
One more question @JuPrgn:
In CO_TIME.h you inlcuded CO_OD.h, why?
First it seems like you have a double inclusion since in CO_Time.c you only include Canopen.h, which already includes CO_OD.h and CO_Time.h
The second thing is, I couldn't find any need for inlusion, since I couldn't find any reference into OD defines or similiar in your code. Did I miss it? Maybe you could me there.

PR #113, opened 2019-10-20
Fix minor typos in comments

Bartimaeus-:
The logic in the comments on lines 231 and 249 should be the opposite of line 178 in order for them to match the logic for the preprocessor directives.
(lines shown below without changes)
https://github.com/CANopenNode/CANopenNode/blob/3f1a6364539708b3b973e0388e789e7d6d66f8ac/CANopen.h#L178

https://github.com/CANopenNode/CANopenNode/blob/3f1a6364539708b3b973e0388e789e7d6d66f8ac/CANopen.h#L231

https://github.com/CANopenNode/CANopenNode/blob/3f1a6364539708b3b973e0388e789e7d6d66f8ac/CANopen.h#L249

### DIFF
--- a/CANopen.h
+++ b/CANopen.h
@@ -228,7 +228,7 @@ CO_ReturnError_t CO_CANopenInit(
         uint8_t                 nodeId);
 
 
-#else /* CO_NO_LSS_SERVER == 1 */
+#else /* CO_NO_LSS_SERVER != 1 */
 /**
  * Initialize CANopen stack.
  *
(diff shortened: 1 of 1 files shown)


martinwag:
This is the same on a few other locations. It's intented this way to easily find/know what the opening statement is, not to show the logic of the else/endif.

martinwag:
I think this will be covered by https://github.com/CANopenNode/CANopenNode/issues/117

PR #116, opened 2019-11-10
Use a void pointer for the CAN device address

odesenfans:
Previously, the address of the CAN device (`CANbaseAddress`)
was specified as a signed or unsigned integer of variable size.
This made the ABI specific for each driver and is not generic
enough for some use cases.

Replaced all usages by a void pointer. Renamed `CANbaseAddress`
to `CANdevicePtr`.
Refactored the code where it needed to be (some comparisons to -1).

### DIFF
--- a/CANopen.c
+++ b/CANopen.c
@@ -391,17 +391,17 @@ CO_ReturnError_t CO_new(void)
 
 /******************************************************************************/
 CO_ReturnError_t CO_CANinit(
-        void                   *CANdevicePtr,
+        void                   *CANdriverState,
         uint16_t                bitRate)
 {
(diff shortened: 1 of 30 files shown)


odesenfans:
I could not test nor even try to compile on most platforms. Additional checks are welcome!

wilkinsw:
I'll aim to this next week on my platform (Kinetis MCU w/ FlexCAN)

odesenfans:
I will just rename the variable, I looked into lwIP to see how they did all that (great lib, there's definitely some inspiration to take from there btw) and they used `state` for the pointer to the private driver state.

PR #120, opened 2019-11-11
Fix bool equality checks

odesenfans:
Removed comparisons of boolean values to `false` and `true`. Evaluating
them is enough.

### DIFF
--- a/stack/CO_LSSslave.c
+++ b/stack/CO_LSSslave.c
@@ -212,7 +212,7 @@ static void CO_LSSslave_serviceConfig(
                 /* Store "pending" to "persistent" */
                 result = LSSslave->pFunctLSScfgStore(LSSslave->functLSScfgStore,
                     LSSslave->pendingNodeID, LSSslave->pendingBitRate);
-                if (result != true) {
+                if (!result) {
                     errorCode = CO_LSS_CFG_STORE_FAILED;
                 }
             }
@@ -331,7 +331,7 @@ static void CO_LSSslave_serviceIdent(
             }
         }
     }
-    if (ack == true) {
+    if (ack) {
         LSSslave->TXbuff->data[0] = CO_LSS_IDENT_SLAVE;
         CO_memset(&LSSslave->TXbuff->data[1], 0, 7);
         CO_CANsend(LSSslave->CANdevTx, LSSslave->TXbuff);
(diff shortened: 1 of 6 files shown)


geoffrey-vl:
Is that considered to be a bad practice?

odesenfans:
@geoffrey-vl yes it's considered bad practice. Nothing major, but it's definitely redundant.

https://softwareengineering.stackexchange.com/questions/12807/make-a-big-deal-out-of-true/12818

martinwag:
Please have a close look at this again. Either I missed something really basic or you are trying to break half of the stack with things like this:

```
--        if(subIndex == 0U  && exception_1003 == false){
++        if(subIndex == 0U  && exception_1003){
```
:-)

I think the discussion on stackoverflow is long enough, but I prefer to have the explicit ==true / ==false when comparing bools in most cases. I think this is more obvious, especially 
```
(bool==false)
over 
(!bool)
```
But I'm open to changing this.

odesenfans:
Thanks for spotting that, I was not careful enough! I started this to replace this:

```
        if (buffer->CANbaseAddress >= 0) {
            return true;
        }
        else {
            return false;
        }
```

Which can be expressed as a one-liner instead. It's generally accepted good practice that you should not compare things to `false` and `true`, but if it's too much of a nitpick no worries! There is more important stuff to change as well :)

geoffrey-vl:
Like mentioned on SO, its a micro optimization. I personally prefer explicit mentioning of true/false, and so do many others. As demonstrated in the PR: it doesn't actually protect you from making big mistakes 😃 
But when maintaining public code I'd encourage going with public accepted conventions and best practices, so I understand the PR and i have no problem dealing with such code.

odesenfans:
It's not even a micro-optimization, the generated code should not change at all from this PR. And we're not in 1980 anymore either so it's not about file size 😄 

It's more about:
a) letting the compiler perform the evaluation of the variable
b) not relying too much on the actual values of `false` and `true`
c) making the code leaner.

No worries if you prefer to keep the existing way either.

geoffrey-vl:
I meant code-readability optimization, not code optimizations  😄

martinwag:
What's the state of this PR? Have you at least triple-checked that this is OK now :-) .

odesenfans:
Hi @martinwag , this can be merged if someone's interested, otherwise if you prefer to keep the current style you can close it. Up to you!

PR #121, opened 2019-11-13
Single project-wide header for driver functions

odesenfans:
Created a single header file (`CO_driver.h`) to declare the functions
of the drivers API. This forces drivers to use a common API.

Previously, each driver declared these functions in its specific
`CO_driver.h` file. This causes two issues:
1. while these functions are used in target-independent code,
   the declarations could be target specific, leading to API/ABI
   compatibility issues.
2. the declarations were duplicated for each driver.

The `CO_driver.h` files for each driver were renamed to
`CO_driver_target.h`. This file is included from `CO_driver.h`.

Added a new header for types: `CO_types.h`. This file currently
holds the definition of `CO_ReturnError_t` that was previously
duplicated for all drivers. Added a few enum values to support
the `neuberger-socketCAN` code.

### DIFF
--- a/CO_driver.h
+++ b/CO_driver.h
@@ -1,7 +1,7 @@
 /**
- * CAN module object for generic microcontroller.
+ * CAN driver functions.
  *
- * This file is a template for other microcontrollers.
+ * Defines functions that must be implemented for each target.
  *
  * @file        CO_driver.h
  * @ingroup     CO_driver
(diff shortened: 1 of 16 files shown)


martinwag:
How closely did you check this? I can take a look at my own driver, but I can't verify all the others...

odesenfans:
I was pretty careful. I apparently have to rebase it now, but I triple-checked the change.

PR #124, opened 2019-11-15
Conditionally compile LED state code

heliochronix:
This implements conditional compilation of LED state tracking code to save cycles and memory for systems that do not have LEDs. It also defines the appropriate macro for drivers that currently make use of this code.


martinwag:
I think getting rid of this stuff when not used is a good idea. However, in my opinion, enabling/disabling should be done inside OD editor alongside the other features and not inside driver header.

heliochronix:
So, inside of the generated CO_OD.h? So the driver code would need to have conditional compilation statements as well?

geoffrey-vl:
Actually Cia 303_3 either specifies using bi-color LED or 2 single color LEDs. Any idea of what is currently supported? From looking at `CO_NMT_Heartbeat.h` I guess only monochrome LEDs is supported.

Regarding how it is implemented. In my opinion the driver that is build for a specific target should optionally implement a way to enable one of both LED implementations. ideally would be both. The application/firmware developer could than decide if he wants to enabled that feature, or not. If the application developer enables it, but the driver does not provide that part, the compiler should generate an error.

heliochronix:
Yeah this sounds like a good idea. I'll edit this PR to update the current two drivers that make use of this code.

heliochronix:
I've implemented those changes. Is there anything further I should do for this?

martinwag:
Wait until I have time to take a look :-)

PR #125, opened 2019-11-15
Fixed warnings associated with using gcc -Wextra

heliochronix:
This simply fixes warnings produced by projects that use GCC with the `-Wextra` flag.


martinwag:
I would prefer to have the "FALLTROUGH" macro definition in one file, not in every one that does a falltrough...

Should we place that inside the driver (or compiler specific) file?

heliochronix:
I agree. I was thinking about where I could centralize that definition, but there's no common header file included by everything other than maybe CO_driver.h. I guess we could do that, or maybe it would be a good idea to create a new common code implementation and header file (byte swapping code and other commonly used support functions could go in there as well maybe? And if we did this, maybe in the future it can check for different compiler provided functionality like GCC's `__builtin_swap` functions in order to optimize certain operations.)

odesenfans:
Ok, ideally what we should do is a central `cc.h` header file that includes a specific header file (`cc_gcc.h`, `cc_clang.h`, etc) that defines all compiler-specific macros. Just have to figure out how we can specify that for custom drivers. If I remember correctly lwIP does this in a clean way.

heliochronix:
Should I close this PR and instead open an issue (maybe with the Enhancement tag or something) to collaborate a solution with you all?

odesenfans:
I'd say that you could create the common header and put the definition there, and improvements will follow. Do you agree @martinwag ?

martinwag:
The thing is, this is endless.  You willl need a compiler **version** specific file at some point, because you are stuck with GCC3.x on your hardware. You will need general and version specific files for the drivers too, as controller manufacturers change their driver API. Don't even think about CPU architecture and OS specific stuff.

I would be fine adding this to the template driver file and implement it in the specific drivers. Most of the drivers are bound to a specific compiler anyway... (Linux -> GCC, ST -> GCC, Microchip -> Microchip, ...). But we can add a compiler definition file also.

blankm:
Basically did the same, just without any additonal headers, just by writing clean code, see #130 
Didn't know others were already on the way to address this.

heliochronix:
I'm going to close this since a couple of the fixes got incorporated through other pull requests, and the fallthrough thing could be figured out later.

PR #126, opened 2019-11-15
Fixed direct use of msg->ident to use driver abstraction function

heliochronix:
This fixes direct use of `msg->ident` to instead use the `CO_CANrxMsg_readIdent()` abstraction function, since not all drivers employ the same `CO_CANrxMsg_t` structure as it is a system-specific struct.

### DIFF
--- a/stack/CO_Emergency.c
+++ b/stack/CO_Emergency.c
@@ -67,7 +67,7 @@ static void CO_EM_receive(void *object, const CO_CANrxMsg_t *msg){
     if(em!=NULL && em->pFunctSignalRx!=NULL){
         CO_memcpySwap2(&errorCode, &msg->data[0]);
         CO_memcpySwap4(&infoCode, &msg->data[4]);
-        em->pFunctSignalRx(msg->ident & 0x07FFU,
+        em->pFunctSignalRx(CO_CANrxMsg_readIdent(msg),
                            errorCode,
                            msg->data[2],
                            msg->data[3],


martinwag:
The CO_CANrxMsg_readIdent() function is not used anywhere else in the code. Have you checked that this is the only place where this is wrong?

heliochronix:
As far as I can tell, the original code base intended this to be the function used to read out the identity of a message. In numerous drivers there's the following comment about the `ident` member of the `CO_CANrxMsg_t` struct. Since this struct is implementation/driver specific, we can't always rely on the `ident` member being used the same way.
```
/**
 * CAN receive message structure as aligned in socketCAN.
 */
typedef struct{
    /** CAN identifier. It must be read through CO_CANrxMsg_readIdent() function. */
    uint16_t            ident;
    uint8_t             DLC ;           /**< Length of CAN message */
    uint8_t             data[8];        /**< 8 data bytes */
}CO_CANrxMsg_t;
```

In addition, the actual function descriptions states the following:
```
/**
 * Read CAN identifier from received message
 *
 * @param rxMsg Pointer to received message
 * @return 11-bit CAN standard identifier.
 */
uint16_t CO_CANrxMsg_readIdent(const CO_CANrxMsg_t *rxMsg);
```

The reason I submitted this change is that this is the first time the field has been used anywhere in the protocol stack code, and it doesn't work with the driver I'm using because the `CO_CANrxMsg_t` struct used to interface with the CAN hardware returns some specially aligned SID/EID field that doesn't work with the new CO_EM_receive code, but does work when using the function that properly parses out that field for use.

PR #127, opened 2019-11-15
Implement real-time function timeout return value

heliochronix:
This implements a return value from real-time functions for use with interrupt driven real-time operating systems. This is implemented similar to the `timerNext_ms` return value from mainline functions, but is meant for use in interrupt driven real-time systems. I also removes the 50ms limit on the `timerNext_ms` return value in order to facilitate extended sleeping operations on interrupt driven low power systems.

The change should be transparent to all currently implemented drivers as it specifically turns the current function calls into wrappers for the return parameter version of the function. For drivers choosing to employ the new function, they call a new function that operates similarly, but with an appropriate variable passed by reference to the function.

A driver that makes use of this on a real-time OS can effectively sleep for large amounts of time until either a CAN message is received that it needs to act on, or a real-time scheduler wakes up the thread based on this returned value.


heliochronix:
There is an open discussion on whether the whole stack should move to a microsecond-based timing system. As it currently stands, the mainline thread functions in the original code base are based on milliseconds, and the real-time thread functions are based on microseconds. I tried to make the return time mechanism consistent with how the millisecond return time mechanism works, while also keeping it consistent with the microsecond time base used in the real-time functions.

heliochronix:
Are there any further thoughts on this mechanism?

As it currently stands, this should be a transparent change for most people. There's already two different time bases used in the existing stack, so at least things are consistent for now with regards to how the RT functions and mainline functions are handled.

We could move the discussion of standardizing the time base between real-time and mainline functions to an issue if no immediate plan to change things is in the works. At least this PR will integrate the real-time return value functionality until we can decide if we want to change the mainline thread function API to be in microseconds.

heliochronix:
Hello again,

I've updated this pull request to not have merge conflicts. Is there anything else preventing this from being merged?

I apologize if I seem to be persistent on this. I'm just hoping to get the RTOS satellite project I'm working on back onto this upstream project, but it's an interrupt-driven tickless RTOS that needs the ability to go into low power mode until either a timer expires for the next event or an interrupt from a received message triggers processing. This provides that functionality, while maintaining backward compatibility with all current code that needs a ticking thread.

martinwag:
Hi,

I didn't merge it because
- I would like to check if we change both time bases to µs. I definitely don't want two different timebases.
- This takes time to estimate and implement. Especially since we want to support 8/16 bit controllers, where we are likely to hit "int" boundary on those devices.
- This takes even more time to check and test that we got it correct in all places
I don't have the time to do that at the moment.

heliochronix:
Hello,

If I can help with this process in any way, please let me know. I've been testing it for a while on our projects. This code change is really quite simple and doesn't really break anything. Perhaps I made it seem more complex than it is.

To clarify, all this PR does is allow a driver developer to call a new process function that simply returns how much time, in μs, is left until the next PDO or SYNC event needs to process. It uses μs because that's what the PDO and SYNC objects use internally.

If I may address the items you've listed:
> I would like to check if we change both time bases to µs. I definitely don't want two different timebases.

I definitely understand this concern. As it stands, the existing code is already using two different time bases for object processed by `CO_process()` and objects processed by `CO_process_SYNC()` and `CO_process_TPDO()`. This new PR simply uses the existing values in the current code base. We'd need to change a lot more than the little bit this PR adds to make that work. I could look into that if you like? Though I think maybe if we do that we'd need to make it a separate issue because the code is already fractured in this regard.

> This takes time to estimate and implement. Especially since we want to support 8/16 bit controllers, where we are likely to hit "int" boundary on those devices.

This does make use of the same `uint32_t` that the existing `uint32_t timeDifference_us` in the code uses. This should force even an 8 or 16 bit MCU to use a uint32_t.

Another thing to consider is that this doesn't increment anything at all, so this PR should have no risk of rolling over anything. It just sets this new return value equal to the smallest difference value used to check against `timeDifference_us` in PDO and SYNC objects, just like the `timerNext_ms` return value in the `CO_process()` function. So if there's an issue with rolling over a uint32_t value when using μs, that issue is already inherent to the existing code base rather than introduced by this PR.

Again, this just returns the smallest time until the next PDO or SYNC object needs to be processed if the driver developer decides they want that value.

> This takes even more time to check and test that we got it correct in all places

If it helps, I've been running this code on some STM32 devices continuously for a few months and have tested the SYNC and PDO processing objects on it.

It's backwards compatible for anything using `CO_process_SYNC()` and `CO_process_TPDO()` (which is all the current drivers). All new functionality is implemented strictly through a new function `CO_process_SYNC_PDO()` that is identical to `CO_process_SYNC()` and `CO_process_TPDO()` save for one extra pass-by-reference parameter that provides the new return value (just like `CO_process()`). The only new thing the existing code will do is check to see if the pass-by-reference value is NULL or not (which it is by default) and move on. If it's not NULL it just sets the value to the next closest `timeDifference_us` required by a PDO or SYNC object. The extra functionality is only there for drivers choosing to know how long it is until the next PDO or SYNC event.

That's all this does really. It just returns the number of μs until the next PDO or SYNC event if a driver developer wants to know that value. Nothing else really.

Is there anything I can do to help?

stack/CO_PDO.c
```diff
@@ -992,6 +997,13 @@ void CO_TPDO_process(
                     TPDO->eventTimer = ((uint32_t) TPDO->TPDOCommPar->eventTimer) * 1000;
                 }
             }
+            if(timerNext_us != NULL){
+                if(TPDO->TPDOCommPar->inhibitTime && *timerNext_us > TPDO->inhibitTimer){
+                    *timerNext_us = (TPDO->inhibitTimer + 9) / 10;
+                }else if(TPDO->TPDOCommPar->eventTimer && *timerNext_us > TPDO->eventTimer){
+                    *timerNext_us = TPDO->eventTimer;
+                }
```
CANopenNode: Probably you mean * instead of + `(TPDO->inhibitTimer + 9) / 10`

I think, you can skip inhibit timer and use only:
```
                if(TPDO->TPDOCommPar->eventTimer && *timerNext_us > TPDO->eventTimer){
                    *timerNext_us = TPDO->eventTimer;
                }
```

CANopen.c
```diff
@@ -761,12 +761,6 @@ CO_NMT_reset_cmd_t CO_process(
         ms50 -= 50;
         CO_NMT_blinkingProcess50ms(CO->NMT);
     }
-    if(timerNext_ms != NULL){
-        if(*timerNext_ms > 50){
-            *timerNext_ms = 50;
-        }
-    }
-
```
CANopenNode: This is not relevant for current commit

stack/CO_SYNC.c
```diff
@@ -337,13 +338,24 @@ uint8_t CO_SYNC_process(
...
         if(SYNC->isProducer && SYNC->periodTime){
+            uint32_t diff;
             if(SYNC->timer >= SYNC->periodTime){
                 if(++SYNC->counter > SYNC->counterOverflowValue) SYNC->counter = 1;
                 SYNC->timer = 0;
                 ret = 1;
                 SYNC->CANrxToggle = SYNC->CANrxToggle ? false : true;
                 SYNC->CANtxBuff->data[0] = SYNC->counter;
                 CO_CANsend(SYNC->CANdevTx, SYNC->CANtxBuff);
+                diff = 0;
```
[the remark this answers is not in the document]
heliochronix: Just to confirm, does the processing function not need to immediately run when the timer has expired?

stack/CO_PDO.c
heliochronix: If there is an inhibit time set, and a COS or application requested send has been queued, do we not want it to go as soon as the inhibit time has expired? I think if we only did the eventTimer it would potentially not send as soon as it was allowed to when the application requests it.

For example, say the eventTimer is 10 seconds, but inhibit time is 1 second and it is configured for a COS send or every 10 seconds max. Then if it sends, the eventTimer is 10 seconds, but a COS occurs 0.25s after the initial send, then instead of waiting another 0.75s (the duration of the inhibit) and sending it would wait another 9.75s.

Or am I not understanding this correctly?

stack/CO_SYNC.c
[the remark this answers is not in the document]
heliochronix: Actually, I guess if it subscribes to the SYNC itself it would immediately trigger the interrupt anyway and process again, so then the next time it would need to run is the next periodTime. I think I understand now.

stack/CO_PDO.c
heliochronix: Also, regarding 

> Probably you mean * instead of + `(TPDO->inhibitTimer + 9) / 10`

This was duplicating the ceiling function implemented in CO_Emergency.c here:

`326         diff = (emInhTime + 9) / 10; /* time difference in ms, always round up */`

So if that is actually wrong, then we should probably change it in that as well correct?

CANopen.c
heliochronix: I'll move it to another PR

heliochronix:
I'm looking into basing it on the `split_driver` branch. I'm not sure if I just merge in `split_driver` and then reopen the PR on that branch or not though?

stack/CO_PDO.c
heliochronix: I updated it to simply do the inhibit time plus 1 to put it past the inhibit window. Is that sufficient? I just realized that the CO_EM_process function apparently operates in 100us increments so I think I see where the confusion lies. Perhaps that's another reason to move to a common microsecond time base haha.

heliochronix:
Closing since I've opened a new PR on #159

PR #128, opened 2019-11-21
change SYNC module to be optional

wilkinsw:
This change will allow CO_NO_SYNC to be zero if the project does not wish to use the sync message.

If CO_NO_SYNC is zero the stack does not need to assign a mailbox to it, which is important if you are using hardware filtering and only have a limited number of mailboxes to work with.  This also means you do not have to define EDS entries 1005-1007.

Changes:
- feature verification no longer asserts on `CO_NO_SYNC == 0`
- `CO_SYNC_t COO_SYNC` is no longer allocated if `CO_NO_SYNC == 0`
- `CO_SYNC_init` will not be called if `CO_NO_SYNC == 0`
- split `CO_process_SYNC_RPDO` into two separate functions.
  - `bool_t CO_process_SYNC`  (returns syncWas)
  - `void CO_process_RPDO` (takes syncWas)
- SYNC is now passed to TPDO on init rather than at every process call.
- Both TPDO and RPDO will now check if SYNC is NULL before dereferencing.

### DIFF
--- a/CANopen.c
+++ b/CANopen.c
@@ -85,7 +85,7 @@
 /* Verify features from CO_OD *************************************************/
     /* generate error, if features are not correctly configured for this project */
     #if        CO_NO_NMT_MASTER                           >  1     \
-            || CO_NO_SYNC                                 != 1     \
+            || CO_NO_SYNC                                 >  1     \
             || CO_NO_EMERGENCY                            != 1     \
             || CO_NO_SDO_SERVER                           == 0     \
             || CO_NO_TIME                                 >  1     \
@@ -143,7 +143,9 @@
     static CO_EM_t              COO_EM;
     static CO_EMpr_t            COO_EMpr;
     static CO_NMT_t             COO_NMT;
+#if CO_NO_SYNC == 1
     static CO_SYNC_t            COO_SYNC;
+#endif
     static CO_TIME_t            COO_TIME;
     static CO_RPDO_t            COO_RPDO[CO_NO_RPDO];
     static CO_TPDO_t            COO_TPDO[CO_NO_TPDO];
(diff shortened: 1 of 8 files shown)


JuPrgn:
Hi, 

This is a great idea to make SYNC optionnal, thank you !

I cannot find `CO_process_SYNC_RPDO()` can you please check if this is missing ?

JuPrgn:
This is working like a charm (SYNC desactivated).

Please note that these files : 
- example\main.c
- PIC32\main_PIC32.c
- socketCAN\CO_Linux_tasks.c

Are still using : 

`syncWas = CO_process_SYNC_RPDO(CO, ...);`

And could be updated with : 
```
syncWas = CO_process_SYNC(CO, ...);
CO_process_RPDO(CO, syncWas);
```

martinwag:
I've had a short look at this. I think disabling things you don't need is a good idea.

We should also conditionally remove the entry from the "Indexes for CANopenNode message objects" Table in CANopen.c as this defines the amount of hardware buffers you need.

JuPrgn:
Please note `free(CO->SYNC);` is not guarded in `CO_delete()`.

wilkinsw:
I'll fix the conflicts and implement those suggestions!

wilkinsw:
@martinwag in response to your comment, CO_RXCAN_NO_MSGS uses CO_NO_SYNC for its calculation, so I don't think touching that block is strictly necessary. We could conditionally define CO_RXCAN_SYNC, but it is only used by CO_SYNC_init() which won't be called if CO_NO_SYNC is 0.

I'd be happy to make that change, but you've mentioned before that you prefer not to touch things unnecessarily to simplify git blame, so I wanted to double check.

wilkinsw:
Actually, no, you're right, it does push the rest of the indexes down. I'll update that.

wilkinsw:
I took another look at it and the indices are incremented using CO_NO_SYNC, so even though CO_RXCAN_SYNC is defined to 1, the next index just becomes (1 + 0).  I don't think we need to change anything there.

martinwag:
Thanks, I missed that.

PR #130, opened 2019-12-05
Heap management functions now mapable according to user needs

blankm:
Malloc, Calloc and Free are now mapped with defines in CO_driver.h, this way the user can map these functions according to the project's needs.
CANopenNode internal function names are now COmalloc, COcalloc, COfree.

### DIFF
--- a/CANopen.c
+++ b/CANopen.c
@@ -161,12 +161,84 @@
 #endif
 #endif
 
+#if CO_NO_LSS_SERVER == 0 /* LSS Server means LSS slave */
+/**
+ * Allocate and initialize memory for CANopen object
+ *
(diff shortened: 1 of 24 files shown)


blankm:
Now also all warnings during compilation are adressed. See latest commit.

martinwag:
I've had a look at this. There are at some other changes I found that are not related to heap management:
- forward declarations in canopen.c
- renaming of *CO to *CO_this
- some changes in switch/case, variable renaming, ...

Please separate those into a pull request for each change, so we can have a discussion on each topic.

blankm:
@martinwag Please see #132 
I will create another PR that will address all the warnings.

PR #132, opened 2019-12-09
Heap management functions now mapable according to user needs

blankm:
Malloc, Calloc and Free are now mapped with defines in CO_driver.h, this way the user can map these functions according to the project's needs.
CANopenNode internal function names are now COmalloc, COcalloc, COfree.


CANopenNode:
This has been merged in #144, except driver files, which will be separated anyway.

PR #133, opened 2019-12-09
NMT Locking

blankm:
NMT object is now being locked when accessed asyncounously
to avoid problems in multitasking/threading environments

### DIFF
--- a/CANopen.c
+++ b/CANopen.c
@@ -174,6 +174,9 @@
         NMTM_txBuff->data[0] = command;
         NMTM_txBuff->data[1] = nodeID;
 
+        /* Protect access to NMT operatingState and resetCommand */
+        CO_LOCK_NMT();
+
         /* Apply NMT command also to this node, if set so. */
         if(nodeID == 0 || nodeID == CO->NMT->nodeId){
             switch(command){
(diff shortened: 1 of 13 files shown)


martinwag:
I think setting this in a multithreading way is a bad idea for master implementation...

However, if you really need this, you have to include the CO_CANsend() function into the lock, as otherwise if you have more than two threads on the lock you will not know which waiting command will be sent last.
If we include the send function in the lock, we introduce the danger of deadlocks. i _don't really_ want to do that.

blankm:
@martinwag 
> However, if you really need this, you have to include the CO_CANsend() function into the lock, as otherwise if you have more than two threads on the lock you will not know which waiting command will be sent last.

Yes, thanks for pointing that out, you are right.

> If we include the send function in the lock, we introduce the danger of deadlocks. i _don't really_ want to do that.

I do not see the danger of deadlocks here, since a thread can only be locked if the CO_LOCK_CAN_SEND Mutex would be taken outside of CO_CANsend. Otherwise it would wait on the NMT_Lock first before even having the chance of blocking inside CO_CANsend?
Maybe I am missing a point here?

CANopenNode:
Agree with @martinwag, closing now.

PR #143, opened 2019-12-20
change TIME module to be optional

wilkinsw:
Add guards to TIME module similar to #128 

Makes defining EDS entries 0x1012 and 0x1013 optional.

@JuPrgn I'd appreciate if you could take a look at this since I do not use TIME for my application.


JuPrgn:
Ok I will give it a try. This is looking fine !

PR #138, opened 2019-12-09
Fix: All Warnings addressed

blankm:
Now finally I am able to provide a PR that addresses all warnings in one unified PR.
Sorry for the multiple PRs before I had to withdraw.

PR #144, opened 2020-01-03
Warnings omitted

lukegluke:
Including:
Shadowing
Missing default cases
Missing cases in switch statements
unused variables
Pointer problems
Function declarations' visibility
Large integer implicitly truncated to unsigned type
Statement may fall through

Most work was done by Michele B.
https://github.com/blankm/CANopenNode.git

### DIFF
--- a/CANopen.c
+++ b/CANopen.c
@@ -165,12 +165,39 @@
 #endif
 #endif
 
+/* These declarations here are needed in the case the switches for the project 
+    change the visibility in the headers in a way that the compiler doesn't see an declaration anymore */
+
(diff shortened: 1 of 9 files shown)


lukegluke:
Instead of https://github.com/CANopenNode/CANopenNode/pull/138

blankm:
We both forgot about LSSslave.c, there are also a lot of warnings in the switch statements.

lukegluke:
I'm new in CANopenNode and now using it as is without LSS inside other project that forbid warnings, so for now I just can only check your work for typos and small corrections. So it is up to you, sorry that I can't help enough, thanks for your work! :)

Issue #149, opened 2020-01-14
Do not use NMTstartup in slave mode, make it optional


lukegluke:
[Here ](https://github.com/CANopenNode/CANopenNode/blob/3d41f20adb2a63737fb7f26cd33a8ae37158b98f/stack/CO_NMT_Heartbeat.c#L224) second bit of NMTstartup (0x1F80 OD element) is used to go into operational state if it zero in all cases.

But according to CANOpen CiA302_v3.0 if device is NOT the NMT Master (0 bit is not set) all other bits have to be ignored.

As for me there is no need in NMTstartup argument of CO_NMT_process at all, but I don't use master, so could someone please take a look how to fix it correctly and do not break master logic.

Also I think 0x1F80 can be optional for slave. So if NMTstartup argument stays for further master functionality, #ifdef OD_NMTStartup really should be added.

CANopenNode:
I verified the standard, we have inconsistency, but also:
- bit0=0 - "CANopen device is not NMT master. The entries of the object 1F81 h shall be ignored. All other bits of object 1F80h shall be ignored with the **exceptions** defined in Table 16."
- bit2=0 - "Shall switch into NMT state Operational in the process NMT startup"
- bit3=1 - "The NMT master shall not start the NMT slaves and the application may start the NMT slaves."
- table16 - exception1, value =  00000000 00000000 00000000 00001000b - "NMT slave that shall enter the NMT state Operational after the NMT state Initialisation autonomously (self starting)"

(DSP302, part2, V4.1.0, 2009)

So, I suggest to fix from:
https://github.com/CANopenNode/CANopenNode/blob/3d41f20adb2a63737fb7f26cd33a8ae37158b98f/stack/CO_NMT_Heartbeat.c#L224

into:
```
            if (NMTstartup == 0x00000008U) NMT->operatingState = CO_NMT_OPERATIONAL;
```

lukegluke:
Ok, but I have only DSP302 V3.2.1 09.04.2003 from here http://overpof.free.fr/schneider/CAN%20&%20CANopen/CANopen/%A9CiA%20CANCANopen%20CD%20V5.1/html/spec3.htm, could you show link where to find DSP302, part2 (?), V4.1.0, 2009.

Your suggestion only about slave, maybe it would be better this?
`if ((NMTstartup == 0x00000008U) || (NMTstartup & 0x05)) NMT->operatingState = CO_NMT_OPERATIONAL;`
[...]

PR #150, opened 2020-01-14
fix NMTstartup handling if device is NOT the NMT Master

lukegluke:
OD_NMTStartup (0x1F80 OD) can be optional

issue #149

### DIFF
--- a/CANopen.c
+++ b/CANopen.c
@@ -824,7 +824,11 @@ CO_NMT_reset_cmd_t CO_process(
             CO->NMT,
             timeDifference_ms,
             OD_producerHeartbeatTime,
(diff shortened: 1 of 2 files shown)


CANopenNode:
Can you fix according to #149

lukegluke:
I make one more commit to fix using OD_NMTStartup for slave too, please merge it too

CANopenNode:
I think, It is OK now.
Files CANopen.h(c) are application specific anyway. After Object Dictionary will be renewed, they will be changed too and probably those files will be moved into "example" directory.

lukegluke:
I told about this commit
https://github.com/lukegluke/CANopenNode/commit/8da2e542f2e1bb4f938b967716cc92f1e9e8ecbf

PR #157, opened 2020-01-22
Added typecast needed for g++

joao404:
g++ throws several errors when compiling because of missed typecasts. Added the needed ones. Files can now be compiled to .o. Compiling of example still not possible because of undefined reference CO_DO.


CANopenNode:
Thank you.
Object dictionary is the oldest part of the project. It is a plan to renew it in the next month.

PR #159, opened 2020-01-27
Implement SYNC and PDO timerNext return value on split-driver dev branch

heliochronix:
This is re-opening #127 on the `split-driver` branch


CANopenNode:
I'm sorry, I'm so nagging. But as @martinwag said, it is no so easy, if we don't want incomplete implementation.

I still don't like the solution with the inhibit timer. Because it will happen, after inhibit timer expires, timerNext_us will be 1 and CO_TPDO_process will just trigger constantly.

I think, the problem is CO_TPDOisCOS() function. However, in #162 I suggested removal of this function and put the option of triggering TPDO into application. 
For now I suggest to somehow set timerNext_us to zero by application, if it writes to OD variable, which could be potentially send with PDO on ChangeOfState.

--------

And one more thing: I would like to pull this enhancement as single commit, preferable into `split-driver` branch or also into `master` branch. I can't do "Rebase and merge", because your commits are to far in history. If I "Squash and merge", Then I get very messed history in gitk and it is not clear, what is changed when.

heliochronix:
No need to apologize, it's not nagging. I'd rather discuss it and get it right!

Looking at the logic in that section, before it sets

`*timerNext_us = TPDO->inhibitTimer + 1;`

it resets TPDO->inhibitTimer via

`TPDO->inhibitTimer = ((uint32_t) TPDO->TPDOCommPar->inhibitTime) * 100;`

if the inhibitTimer has expired and it successfully sends the TPDO. I think the only case in which it would re-run immediately would be if it fails to CO_TPDOsend() immediately, correct? Perhaps I can rework the logic slightly to have it wait for the attempt to send to complete. Does that seem reasonable?

Slightly off topic, but regarding the triggering of a TPDO, that's what the TPDO->sendRequest flag is for, correct?

I can turn all this into a single commit if you like, no problem!

I'll also create a single commit for the PR.

CANopenNode:
I will check this tomorrow.
If you can, pleas make single commit. Multiple consecutive commits are also OK, I think.

I will also make all time bases to microseconds.

heliochronix:
Done

PR #166, opened 2020-02-13
Fixing typos in macro definitions for CO_NO_LSS_*

heliochronix:
Simple edit to correct some macro definitions that had typos. Specifically CO_NO_LSS_MASTER and CO_NO_LSS_SLAVE should be CO_NO_LSS_CLIENT and CO_NO_LSS_SERVER.

Also removed a spurious parenthesis that I think was an accident for CO_NO_TRACE while I was at it.

### DIFF
--- a/CANopen.h
+++ b/CANopen.h
@@ -115,10 +115,10 @@ extern "C" {
 #if CO_NO_SDO_CLIENT != 0
     #include "301/CO_SDOclient.h"
 #endif
-#if CO_NO_LSS_SLAVE != 0
+#if CO_NO_LSS_SERVER != 0
     #include "305/CO_LSSslave.h"
 #endif
-#if CO_NO_LSS_MASTER != 0
+#if CO_NO_LSS_CLIENT != 0
     #include "305/CO_LSSmaster.h"
 #endif
 #if CO_NO_TRACE != 0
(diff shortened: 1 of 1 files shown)


martinwag:
How did this get in? Was originally "CO_NO_LSS_SERVER" / "CO_NO_LSS_SERVER"?

heliochronix:
Looks like commit ae48e57 accidentally added the wrong names.

CANopenNode:
It is my typo.

I renamed some files and I wanted to make things consistent with standard. Standard uses client/server for SDO and master/slave for LSS and NMT.

But I didn't want to change CO_NO_LSS_SERVER/CLIENT yet, because it depends on libedssharp (I think). So I meant to leave this as it was for now, but made a typo with includes. I wanted to discuss this a little when new OD will be actual.

heliochronix:
For what it's worth, the CO_OD.* code is already starting to not be compatible. The #include directives no longer find the header files. I think if you want to change the naming scheme you could just go for it on this branch and it can drive the changes on libedssharp when it gets updated for includes maybe.

CANopenNode:
Actually CO_OD.* code is still perfectly compatible. Except my typo.

heliochronix:
I'll have to check which commit my libedssharp is at. I swear I thought I pulled the latest and it didn't generate `#include "301/CO_SDOclient.h"`. I'll try again

heliochronix:
Yeah, as of [this commit](https://github.com/robincornelius/libedssharp/blob/2e53bbcb7c012805f02c7b9336efc67c743a95b1/libEDSsharp/CanOpenNodeExporter.cs#L741) it still seems to show it's using the old headers. I think you updated your examples but I was saying libedssharp already needs updating so there's probably no reason to let that hold you back from mass updating the macros.

CANopenNode:
I try not mix to much two steps: split-driver and new-OD.

heliochronix:
Agreed, that's a good idea.

CANopenNode:
I fixed my typos with last commit, where I also added some more documentation, especially ChangeLog. Confusion with server/slave will be solved with newOD.

For now headers in CO_OD.c will have to be edited manually. It is mentioned in readme. Othervise I think, libedssharp stays comaptibile.

Thank you for your corrections, you are very up-to-date.

Issue #39, opened 2017-05-23
SDO communication after "SDO block upload end" is dropped


martinwag:
When doing a SDO block upload and **immediately** after that starting another SDO request, this request is dropped.
This happens e.g. in the CANopen conformance test tool when checking SDO test 19 followed by 21. This is equivalent to the following linux commands

```
#do SDO upload block
cansend can0 67f#a0.00.10.00.14.00.00.00
sleep 0.1
cansend can0 67f#a3.00.00.00.00.00.00.00
sleep 0.1
cansend can0 67f#a2.01.14.00.00.00.00.00
sleep 0.1
#SDO block upload end
cansend can0 67f#a1.00.00.00.00.00.00.00
#immediately start next SDO communication
cansend can0 67f#C2.17.10.00.02.00.00.00
```

I've narrowed this down to the CO_SDO_receive() function. This receives the new request before the "SDO block upload end" message is processed, and therefore just drops it. 
When looking at the spec, I can't find anything that would speak against having this sequence. I also think this is a realistic scenario for the client to just start the next request after one has finished.

From looking at the code, I have no idea how to fix this easily, at least for upstream...

I've also had a look at the other SDO transfer methods. They are not affected as the end message is always from server to client.

reference: CiA301, 7.2.4.3.12 Protocol SDO block upload

martinwag:
This dirty hack makes it work when using FreeRTOS Threads for receiving and processing.
```
static void CO_SDO_receive(void *object, const CO_CANrxMsg_t *msg){
    CO_SDO_t *SDO;

    SDO = (CO_SDO_t*)object;   /* this is the correct pointer type of the first argument */

    extern void vTaskDelay( const uint32_t xTicksToDelay );
    while (SDO->CANrxNew) {
      vTaskDelay(1);
    }
...
```

CANopenNode:
Block transfer is quite complicated and large part of code and it needs fast response. SDO->pFunctSignal(); should be used with RTOS. 
However, problem you describe still exists and one solution is, as you describe.

I think, I should make SDO block transfer as optional feature in CANopenNode. It should be possible to not include it's code. I think, it is not very widely used and it may add unnecessary memory consumption.

Another thing to consider is block transfer in future editions of the standard. I checked the latest 301 proposal, which includes also CAN FD messages (64 bytes long). SDO transfer is quite different there and block transfer is not used.

For now, I think, careful integration with RTOS is best practice. I added a note into the stack.

[...]

PR #174, opened 2020-04-15
add fifo queue for SDO received messages to fix possible drops

lukegluke:
Previously when doing a SDO block upload and immediately after that starting another SDO request, this request was dropped. Especially if processing function has slow response.

fifo queue size is 2 by default. To increase performance a little, one can redifine CO_SDO_RX_DATA_SIZE to 1 in CO_driver_target.h, if block transfers are not used

issue #39

### DIFF
--- a/stack/CO_SDO.c
+++ b/stack/CO_SDO.c
@@ -160,6 +160,19 @@ void CO_memcpySwap8(void* dest, const void* src){
 }
 #endif
 
+static void CO_SDO_receive_done(CO_SDO_t *SDO){
+#if CO_SDO_RX_DATA_SIZE > 1
+    uint8_t rcv = SDO->CANrxRcv;
+    uint8_t newRcv = rcv;
+
+    if (++newRcv >= CO_SDO_RX_DATA_SIZE)
+        newRcv = 0;
+    SDO->CANrxRcv = newRcv;
+    SET_CANrxNew(SDO->CANrxNew[rcv]);
+#else
+    SET_CANrxNew(SDO->CANrxNew[0]);
+#endif
+}
 
 /*
  * Read received message from CAN module.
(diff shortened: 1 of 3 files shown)


lukegluke:
Hi @CANopenNode!

I've tested this fix in my environment and everything is ok, including case CO_SDO_RX_DATA_SIZE=1 - driver works as before. But I do not use timerNext_ms and CANrxMemoryBarrier, please take a look maybe I've broken some functionality in this cases.
It will be nice if @martinwag you could check this too. Thanks in advanced!

martinwag:
Hi,

I've had a look, I think it should not break anything. Sync is still at the same position in code, so that should work. Timer is set to zero to request another processing run, that should work also. Other stuff is mostly to adopt for FIFO, that should be fine if it works with you. 
Please be aware that I did not run the code.

lukegluke:
Thanks @martinwag for feedback.

Do not pull this, I need to test it more, probably encountered with some issues.

lukegluke:
I get lucky to have several issues in external SDO client software, so I get encountered on bug: fifo process index overrun on timeout in sub-block transfers.

Now I get fixed it.
Also I think it is right decision to skip all received message in queue in SDO_abort. Do you agree?

I've tested this with consistent several block downloads and uploads (about 21 Kb in all), some of them alternate with expedite transfers. Also I've test CO_SDO_RX_DATA_SIZE=1 case, and some SDO errors so SDO_abort is called. Looks like everything works fine now.

martinwag:
I think you can drop the messages on abort. However, this should not make any difference as all messages until next _SDO down/up initiate_ need to be dropped anyway (including those that are still on the wire).

CANopenNode:
I just merged split-driver into master branch. But I can still merge these fixes into old-master branch, which I also created.

In my plans for new object dictionary I will probably somehow rewrite SDOserver and will certainly include those fixes also into new code.

lukegluke:
> But I can still merge these fixes into old-master branch, which I also created.
> In my plans for new object dictionary I will probably somehow rewrite SDOserver and will certainly include those fixes also into new code.

Ok. Thanks for work!

CANopenNode:
This is merged as is into branch old-master.
Thank you.

CANopenNode:
Actually it is inside branch `v1.3-master`.

lukegluke:
Thanks.
@CANopenNode by the way, what do you think about https://github.com/lukegluke/CANopenNode/commit/2c0b8473727f874d53f03e62f2feeceb10601733 ?

CANopenNode:
I plan to solve those issues with new Object Dictionary, which I will start soon.

lukegluke:
By the way, @CANopenNode in new master you can set CO_SDO_RX_DATA_SIZE to 2 only if block transfers are defined.

PR #175, opened 2020-04-22
Replace all memcpy/memset-like functions with standard <string.h> functions

FreddieChopin:
1. Replace all uses of `CO_memcpy()` with `memcpy()`. Both these functions are already used in the code, so no need to have them both.
2. Same as above, but with `CO_memset()` -> `memset()`.
3. Reimplement `CO_getUint16()`, `CO_getUint32()`, `CO_setUint16()` and `CO_setUint32()` with standard use od `memcpy()`, which doesn't require any union.
4. For little-endian systems `CO_memcpySwap2()`, `CO_memcpySwap4()` and `CO_memcpySwap8()` actually do nothing, so just replace them with `memcpy()`.
5. Remove `CO_bytes_t` type, which was used by previous implementation of `CO_{get,set}Uint{16,32}()`.


martinwag:
Hi,

you can even go a step further and get rid of memcpy for the getters/setters like this
```
uint32_t CO_getUint32(const uint8_t data[]){
    return *((uint32_t*)data);
}
```

@CANopenNode what do you think? I'm in favour of this because
- a modern compiler might just optimize away memcpy(), ... entirely for our use cases
- someone new to the code doesn't have to check what CO_memcpy() is, only to find out that it's our own reimplementation of memcpy().
- less code is good :-)

FreddieChopin:
> you can even go a step further and get rid of memcpy for the getters/setters like this...

You cannot, unless you can guarantee that the pointers are always properly aligned. On some 32-bit architectures (ARM Cortex-M3) unaligned access works fine, but on others (ARM Cortex-M0) it  ends up in a core exception ( http://infocenter.arm.com/help/index.jsp?topic=/com.arm.doc.dui0497a/BABFAIGG.html ).

Side note - previous getters/setters were using "undefined behaviour", memcpy is perfectly defined, casting pointers "up" is (in majority of cases) "undefined behaviour" again. This may be important for people who have to deal with nonsenses like MISRA C and similar.

FreddieChopin:
> a modern compiler might just optimize away memcpy(), ... entirely for our use cases

BTW - if we want optimization, than it might be a good idea to implement some of these functions (at least getters and setters) as inline (preferred) or as macros. If CANopenNode's minimum C standard would be declared as C99 (a good idea in my opinion - it's 2020 already [; ), then nice inline functions with type checking would be a very good approach!

CANopenNode:
I agree with the changes.
It is just a little problem here. I'm intensively working on split-driver branch. I plan to merge it to master soon, but new branch is quite reorganized. I don't like to maintain two different branches...

What do you think about new branch?

FreddieChopin:
If you want I could also provide another pull request which would implement the same changes in the split-driver branch. Or I can provide this next pull request after your branch gets merged to master. But in each case I think it would be beneficial to have it merged in the current master branch as-is, unless you plan to merge your branch really soon.

CANopenNode:
Ok, then I think, it is time to merge split-driver branch into master. It is tested to some extent, but I can't guarantee fully for it. On the other side, if I wait too long, there will be just a mess.
However, I just merged the split-driver into master. There is also an old-master branch if someone will need it.

And please, don't blame me too hard, I may have some bad programming habits. I'm not so professional programmer and for some principles I'm not so sure. I'm learning.

I agree with more modern approach, inline functions should not be a problem. C99 as minimum should be OK.

And one more note: I'm currently working on SDO code, actually SDO client in connection with CiA309-3-CAN-ascii-gateway. (I is hard, I didn't believe @martinwag, he was right). However, now I have practically rewritten SDOclient.h. It is much more clear and manageable now with all the block transfer, etc. It is not published yet, because it is not tested.
My current effort is a good foundation for next step, new OD, where SDO server will also have to be rewritten with the necessary getters/setters, etc.

martinwag:
@CANopenNode I've written a CANopen manager / ASCII master application on top of CANopenNode, if you have questions feel free to ask. But please, do separate repo for manager/master, this is **definitely** a project on it's own.

CANopenNode:
I certainly don't want CANopenNode to become fat or interlaced. Actually I'm trying to keep it configurable also for thin configuration. For example enable only expedited SDO transfer, etc.

There are two options:
1. One project, some optional modules, like 302(CANopen manager), 309 (gateway), but no device profiles or project or microcontroller specific code. Everything well organized and independent as possible. Only modules independent from microcontroller included (except basic Linux driver)
2. Separate projects for each module, like CiA309, CiA302, etc.

I prefer first option, because everything is nice on one place and everybody use only, what he needs. There may be some extra code, which can be safely ignored. Otherwise, I'm afraid, things can become too partitioned. For example, SDO client should not be part of the basic stack.

Sorry for off-topic.

FreddieChopin:
I've updated the branch - it is now based on current master. While at it I noticed several bugs in calls to `CO_memset()`/`memset()`, which I also fixed here.

PR #178, opened 2020-05-04
Add CO_..._initCallbackPre() for RPDO and SYNC

FreddieChopin:
Add CO_RPDO_initCallbackPre() and CO_SYNC_initCallbackPre(), which are functionally identical to functions for SDO, NMT, HB and so on. With the addition of these two new functions, a callback invoked after reception of SYNC or RPDO can immediately wake the thread that processes these messages. Moreover - it is now much easier to handle CANopenNode in _one_ thread (instead of two - one for CO_process() and another one for periodic processing of SYNC, RPDO and TPDO). With these callbacks everything can be easily and efficiently handled in a single thread. Such change will drastically reduce RAM consumption (1 thread instead of 2) and also OS overhead (less context switches), at zero cost - due to callbacks the default value of timerNext_us can be extremely big (say UINT32_MAX), but the latency will be nearly zero, there will also be nearaly zero useless runs of this thread.

### DIFF
--- a/301/CO_PDO.c
+++ b/301/CO_PDO.c
@@ -24,6 +24,8 @@
  */
 
 
+#include <string.h>
+
 #include "301/CO_driver.h"
 #include "301/CO_SDOserver.h"
 #include "301/CO_Emergency.h"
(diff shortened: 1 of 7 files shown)


FreddieChopin:
Two comments (;

1. It is also possible to add such callbacks for TIME and for LSS slave, but I first want to know whether this change is acceptable before creating another pull request for the remaining two types. Also SYNC and RPDO are much more important, as with these two callbacks you can easily get rid of the "periodic thread" and handle everything **efficiently** and **quickly** in single thread. This is of course also possible with current master, but in that case it is not that efficient and there is still a latency in reacting to SYNC or RPDO.

2. I did not modify the code for SocketCan driver to actually use this new features. I could try to do that, but I think this has to wait until CANopenSocket repo is updated to use the new CANopenNode.

CANopenNode:
I was also thinking about single thread in Linux. Actually i would like to keep both options:

1. Two threads, where one is real-time (1ms) timer thread, which is now the only option. It is possible to add efficient custom program into realtime thread and build a powerful controller.

2. Single thread for more efficient operation. There is also a timerNext_us variable, which can be used for larger timer interval (100ms or more instead of 1ms). It further reduces processor consumption. timerNext_us was recently implemented for whole stack, see #163.

---
There is also CANopen-ascii gateway (command interface) in CANopenSocket, which uses own thread, because it is blocking function. But, as I said, i have rewritten that part and now it is non-blocking, so ready for single thread operation. It will soon be published.

CANopenNode already contains Linux driver, main() and it compiles to fully featured application. It is a replacement for canopend from CANopenSocket. I will add command interface as optional, microcontroller independent module into CANopenNode. Command interface will also be included from CO_Linux_threads.c file.

You can update socketCAN Linux driver in CANopenNode with one threaded operation, if you are so kind.

---
I'm merging your changes, thank you.

PR #193, opened 2020-06-02
CO_trace: fix warning "cast increases required alignment of target type"

lukegluke:
This extends #144, commit f2283e3ca98650b321ae51a2d4e69e21dd45aa32


lukegluke:
With last commit f60f03c7 in v1.3-master that revert "guard CO_trace to be optional" encountered on not fixed warnings like we did with #144.

But I wonder, why this reverting f60f03c7 was done?

geoffrey-vl:
The reverting was done because of an include loop issue: https://github.com/CANopenNode/CANopenSocket/pull/17#issuecomment-634585107

Issue #192, opened 2020-06-01
Change rx, tx buffers ident on the run for additional SDO


lukegluke:
Hi @CANopenNode!

For now CANopenNode supports multiply SDO servers, but only with predefined  parameters. One must set client to server and server to client COB-IDs in 0x1201 (and up to 0x127F according to number of servers) before SDO initialization otherwise additional SDO servers remain invalid.

I would like to implement handling dynamically changes of parameters of additionally supported SDO servers. CO_CANtxBufferInit returns pointer to CO_CANtx_t * buffer and that's good and useful in any ways, and particulary for my case I could use it to set different ident (for example adding CO_CANrxBufferSetIdent to CO_driver).
But I wonder why CO_CANrxBufferInit doesn't return a pointer to buffer? It's just a legacy, because it was needed anywhere?
Changing CO_CANrxBufferInit is quite a big thing that brakes quite a lot, you will not merge such changes (in old master)?

CANopenNode:
There is actually no serious reason, why CO_CANrxBufferInit doesn't return a pointer to buffer.

I think, it is not such a big thing to change that. Stack is now in the middle of the change between major releases anyway.

Currently CO_CANrxBufferInit returns CO_ReturnError_t, but the only information needed is, if init succeeds or fails. So, if it returns NULL pointer in case of failure, it is just fine. It will be necessary to update multiple files in CANopenNode (only).

If you will change old-master, I will change master too.

lukegluke:
Ok, good! I will try my approach.

lukegluke:
In fact returned CO_ReturnError_t was used only in PDO and SDO.
I've implement dynamic SDO, I make pull request let's continue there.

PR #194, opened 2020-06-02
Handle dynamic changes of additional SDO servers parameters

lukegluke:
I've tested it on hardware with one and then two additional servers (that is 3 including default one).
It works fine, I can config and reset servers from other SDO servers, and even can turn server from itself (it just stops working on writing COB-ID with invalid flag).

But let's discuss this a little before merging.

A make two separate commits.
First one changes CO_CANrxBufferInit to return pointer to receive buffer as I mentioned #192.
I encountered that only PDO checks CO_CANrxBufferInit result on error. Probably we should extend this commit as something separated and add checking to other places where CO_CANrxBufferInit is used?

Second commit related to dynamic changes of additional SDO servers parameters.
CO_SDO_ODF_12xx in CANopen reveal some internal SDO logic, but I think it is only suitable place, because CO_SDO related only for single SDO server and doesn't now anything about others.
Is CO_SDO_AB_DATA_OD the appropriate error to return if something wrong with index? Or CO_SDO_AB_NO_RESOURCE maybe suits better?

Also please review style and instructions-comments. Probably you may want to correct and extend them, and also maybe add feature explanations/instructions in readme, please fell totally free to do this.

### DIFF
--- a/CANopen.c
+++ b/CANopen.c
@@ -526,34 +526,16 @@ CO_ReturnError_t CO_CANopenInit(
         return CO_ERROR_PARAMETERS;
     }
 
-    for (i=0; i<CO_NO_SDO_SERVER; i++)
-    {
-        uint32_t COB_IDClientToServer;
-        uint32_t COB_IDServerToClient;
-        if(i==0){
-            /*Default SDO server must be located at first index*/
-            COB_IDClientToServer = CO_CAN_ID_RSDO + nodeId;
-            COB_IDServerToClient = CO_CAN_ID_TSDO + nodeId;
-        }else{
-            COB_IDClientToServer = OD_SDOServerParameter[i].COB_IDClientToServer;
-            COB_IDServerToClient = OD_SDOServerParameter[i].COB_IDServerToClient;
-        }
-
-        err = CO_SDO_init(
-                CO->SDO[i],
-                COB_IDClientToServer,
-                COB_IDServerToClient,
-                OD_H1200_SDO_SERVER_PARAM+i,
-                i==0 ? 0 : CO->SDO[0],
-               &CO_OD[0],
-                CO_OD_NoOfElements,
-                CO_SDO_ODExtensions,
-                nodeId,
-                CO->CANmodule[0],
-                CO_RXCAN_SDO_SRV+i,
-                CO->CANmodule[0],
-                CO_TXCAN_SDO_SRV+i);
-    }
+    err = CO_SDO_init(
+            CO->SDO,
+           &CO_OD[0],
+            CO_OD_NoOfElements,
+            CO_SDO_ODExtensions,
+            nodeId,
+            CO->CANmodule[0],
+            CO_RXCAN_SDO_SRV,
+            CO->CANmodule[0],
+            CO_TXCAN_SDO_SRV);
 
     if(err){return err;}
 
--- a/stack/CO_SDO.c
+++ b/stack/CO_SDO.c
@@ -24,7 +24,7 @@
  */
 
 
-#include "CO_driver.h"
+#include "CANopen.h"
 #include "CO_SDO.h"
 #include "crc16-ccitt.h"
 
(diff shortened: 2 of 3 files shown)


CANopenNode:
I'm trying to transfer your commit to master branch too and I found a problem in your principle.

The problem is with setting CAN filters and is present only in some microcontrollers. (for example see https://github.com/CANopenNode/CANopenNode/blob/master/socketCAN/CO_driver.c#L474)

If we need to change CAN identifier for the received message, we have to call whole CO_CANrxBufferInit() function. By changing identifier, it is also necessary to reconfigure CAN hardware filters. And for that we need reference to CANdevRx and CANdevRxIdx. In that case your functions CO_CANrxBufferSetIdent() and CO_CANtxBufferSetIdent() won't work with more advanced microcontrollers and Linux.

It is possible to change CAN identifier in normal CANopen operation. Actually this is implemented in SDO clinet, HB consumer, RPDO and SYNC. The point is, you have to remember (put in CO_SDO_t) also CANdevRx and CANdevRxIdx.

Please put CO_SDO_ODF_12xx() or similar rather to CO_SDO.c file. For reference look into CO_HBconsumer.c file. It is no problem, if you add additional arguments to CO_SDO_init().

So it does not make sense to change return value of CO_CANrxBufferInit() function. I'm sorry for my wrong advice in #192

lukegluke:
> The problem is with setting CAN filters and is present only in some microcontrollers

Yes, this is what I was afraid of. I don't have opportunity to work with hardware filters.

Thanks for feedback, I will check your suggestions and try to correct code latter, hope you can support with checking logic to work correctly with hardware filters.
But one current thought:

> Please put CO_SDO_ODF_12xx() or similar rather to CO_SDO.c file.

I probably can move CO_SDO_ODF_12xx to CO_SDO as "public" function, but CO_OD_configure will still be in CANOpen.h I suppose, because otherwise it will demand to put into CO_SDO_init and CO_SDO some upper level logic with several servers, I don't think it is good and I tried to avoid this.
Also I would like to make CO_SDO_ODF_12xx optionally - only when there are several SDO (CO_NO_SDO_SERVER > 1).

CANopenNode:
Now latest master contains all updates, that was applied to v1.3-master.
pr #171 was included some time ago and pr #174 was included just now.

lukegluke:
Thank you!

lukegluke:
Hi @CANopenNode 

Finally got around to correct this PR. Sorry, I'm still on old master.

I've start with feedback approach, but got messed and ugly code. So in the end I stopped on configuring all SDO in SDO.c instead of CANopen.c. Previously I thought that with this I take upper level logic into CO_SDO_init(), but in fact SDO servers a quite coupled. Mixing logic CANopen.c and SDO.c with help of feedbacks was much worse.

So I've just broke one principle I mentioned and now CO_SDO is related to all SDOs servers (in fact just in init and CO_SDO_ODF_12xx handling) and it is totally ok as for me. Also bonus - CANopen.c get simpler as you prefer :)

I've tested it on hardware with zero, one and then two additional servers (that is 3 including default one).

By the way some little thought: previous code have one strange potential for future - additional (not the default one) SDO server could have its own separate OD. Can there be at least one such case sometimes? I can't imagine this and anyway it will be quite a challenge to make everything right.

And about change return value of CO_CANrxBufferInit() function. It doesn't really matter now, but don't you think it would be better someday to bring it into line with tx maybe for future use? It looks more flexible - for error one can always check returned pointer for null.

p.s. Could someone explain  the meaning of Node-ID of the SDO client - subindex 03h of OD 1200h to 127Fh in spec? A filter - only this Node can access this?

lukegluke:
@CANopenNode it's better to merge #204 before this.

CANopenNode:
I would like to wait with this pr, because I'm working on renewed SDO server now. Just to avoid mess with different versions.

p.s. I think, Node-ID parameter  of the SDO client - subindex 03h is an alternative for easier configuration. If 'Node-ID of the SDO server' is written, then 'COB-ID client to server' and 'COB-ID server to client' should be configured automatically. But I'm not sure.

lukegluke:
> I would like to wait with this pr, because I'm working on renewed SDO server now. Just to avoid mess with different versions.

Ok, it's up to you.

> p.s. I think, Node-ID parameter of the SDO client - subindex 03h is an alternative for easier configuration. If 'Node-ID of the SDO server' is written, then 'COB-ID client to server' and 'COB-ID server to client' should be configured automatically. But I'm not sure.

In our project we indeed set 'COB-ID client to server' and 'COB-ID server to client' for extra SDO to be corresponding to default SDO of another (kinda "virtual") Node with another ID, and your suggestion sounds good for me. And it's totally ok for particular CANopen net.
But in fact it would be against the letter of the 301 specification. You can find that in description of CAN-ID field for index 1201h to 127Fh there is a note: _see clause 7.3.5_. 

> 7.3.5 Restricted CAN-IDs
> Any CAN-ID listed in Table 40 is of restricted use. Such a restricted CAN-ID shall not be used as a
> CAN-ID by any configurable communication object, neither for SYNC, TIME, EMCY, PDO, and SDO.
изображение

So IDs of default SDO of other nodes are restricted in general case.

CANopenNode:
Sorry, I was looking into 'Node-ID of the SDO server' from 1280 object.

I can only guess about 'Node-ID of the SDO client' from 1200-127F objects. This is optional parameter and it is not used by CANopenNode. I don't know, why this parameter can be useful, it may be there only for optional information.

I checked the standard CiA302-5 (Additional application layer functions, Part 5: SDO manager), there is some explanation:

> CANopen offers a communication mechanism between CANopen devices via SDO. These
> communication channels are always established between two CANopen devices. For
> accessing a CANopen device the first time at least one SDO channel per CANopen device is
> required. This is called the default SDO channel. Only the SDO manager is allowed to use
> that SDO channel.
> Simple networks may use SDO channels that are pre-configured between a pair of CANopen
> devices. This allows direct communication between CANopen devices without the need of a
> SDO manager.
> CiA302-5 defines mechanisms that may be used for plug-n-play networks, without the
> requirement of a pre-configuration of the network. These mechanisms may be used to
> establish dynamic SDO connections between CANopen devices. These mechanisms require a
> specific CANopen device in the network that is able to handle the dynamic request; this
> CANopen device is called SDO manager. These mechanisms require CANopen devices that
> are able to perform a dynamic request; these CANopen devices are called SDO Requesting
> Device (SRD).

I think, you are right about '7.3.5 Restricted CAN-IDs'. If SDO channels are manually pre-configured, then they should not use default CAN ids.

lukegluke:
> I think, you are right about '7.3.5 Restricted CAN-IDs'. If SDO channels are manually pre-configured, then they should not use default CAN ids.

Just in case, if you thinking about adding restriction checks into CO_SDO_ODF_12xx, please envelope them in ifdef, so advanced users could violate specification a little if they now what they are doing :)

lukegluke:
> I'm working on renewed SDO server now.

Hi, Janez, how it is going on with this?

CANopenNode:
I just pushed new `301/CO_ODinterface.h/c`, SDO server and everything else is coming next.

CANopenNode:
newOD is now near to be finished, see https://github.com/CANopenNode/CANopenNode/issues/162#issuecomment-751724920

newOD addresses all issues we had with old one. I think it is quite flexible. So I think, it is best to keep current OD and SDO server as is.

PR #195, opened 2020-06-04
remove the need to specify the number OD entries

dergentk:
The number of entries in the OD was specified in CO_OD.h. This is fine as long
as the CO_OD.h and CO_OD.c files are generated automatically, but is very error
prone in cases where these files are maintained manually.

Here we remove the need to specify that count up front, while still staying
compatible with existing code.


CANopenNode:
I'm now just preparing to start with New Object Dictionary approach, as described in #106 
So I try to make as little as possible changes to this part now. I don't like to pull your request at this moment, sorry.

dergentk:
No problem. Thanks for letting me know.

PR #204, opened 2020-06-30
SDO: fix to always check if buffer is empty before freeing it

lukegluke:
This is critical fix for PR #174 (issue #39)

In 5967fc607c502bb52009e8ecaa7c565f5f55f908 I missed checking isNew also in `if(!NMTisPreOrOperational){}`. Without this process pointer could overrun receive pointer on receiving NMT stop command during active SDO communication - this totally breaks all further proper working.

Now checking, if buffer is empty before freeing it, is always done inside CO_SDO_process_done for sure.


CANopenNode:
This is also fixed in master branch.

Issue #173, opened 2020-04-14
Multiple instances?


FreddieChopin:
Hello!

In some CANopen profiles there is a need to have many independent CANopen instances on one device. For example "CANopen Application Profile for Lift Control Systems" (DSP417) describes several configurations where something like that is required. most important ones:
- devices working as gateways/bridges/extenders,
- central controller devices which need to access two/three separate networks (for example one network is for all "call" buttons next to lift doors, second network is for communication with the lift itself, third network for controlling motor drives).
![Screenshot from 2020-04-14 21-49-44](https://user-images.githubusercontent.com/1139038/79267704-e2e55d80-7e99-11ea-8550-5af9f484096a.png)
![Screenshot from 2020-04-14 21-50-09](https://user-images.githubusercontent.com/1139038/79267772-fdb7d200-7e99-11ea-80c8-55aacfe0e941.png)
![Screenshot from 2020-04-14 21-50-29](https://user-images.githubusercontent.com/1139038/79267784-027c8600-7e9a-11ea-8676-72faaedfc080.png)

I'm pasting this to show what's the real-world use case for a multi-instance scenario.

Let's say I'm mostly interested in a scenario with two completely separate object dictionaries and two separate CAN interfaces in a single device. Let's also say I'm using a MCU with flat address space, not an embedded Linux with processes and virtual addressing. Am I right to assume that such things cannot be done with CANopenNode at the moment? Or maybe doing something like this would be possible, but very awkward? Maybe it just requires some minor edits in the source files?

Thanks in advance for any clarification!

CANopenNode:
Hello,
The problem is Object Dictionary, where currently are fixed global variables. This is oldest part of CANopenNode and is planned to be rewritten in next few months. For now it could be possible to manualy rename global variables in second object dictionary.
Next thing are CANopen.h/c files, which are fixed to one CANopen device. For complex configuration they have to be customized.
Other CANopen objects are independent from Object Dictionary and from CANopen.h/c files. I think, there should not be much effort to implement two CANopen interfaces.

Please note also, there is a branch split-driver, which will soon be merged into master and is recommended for new designs.

martinwag:
Hi,

I think the most difficult problem with this is to get the driver right, as you now have multiple instances using one driver. This means you have to handle calls independently to some degree. This also means you have to analyse received messages to select a device instance (or all of them) a message is sent to.

FreddieChopin:
[...]

PR #206, opened 2020-07-02
Change array of pointers to a pointer + size

FreddieChopin:
This is a first stage of an effort to making CANopenNode support multiple instances #173 . If the general direction I'm heading at with these changes is acceptable, there will be more pull requests.

The `CO_t` struct has multiple fields which are arrays of pointers, of fixed size (`CO_NO_...`). This serves no purpose, just makes the code more complex. Turn these fields into normal pointer to the first element of an array, additionally add a field which holds current size of the array assigned to the pointer. This way the global constants/defines (`CO_NO_...`, like `CO_NO_TPDO`) are used in smaller number of places (basically only in the `CO_new()`), making some of the functions (more or completely) independent from the global state.

These changes may break some user code, however the fixes are very trivial:
- replace `CO->CANmodule[0]` with `CO->CANmodule`,
- for all `..._initCallbackPre()` add `&` to the first argument, e.g. `CO->SDO[0]` ->`&CO->SDO[0]` (`CO_SDO_initCallbackPre(CO->SDO[0], NULL, threadMainWait_callback)` -> `CO_SDO_initCallbackPre(&CO->SDO[0], NULL, threadMainWait_callback)`),
- as above for `CO_trace_process()` - `CO_trace_process(CO->trace[i], *CO_time.epochTimeOffsetMs)` -> `CO_trace_process(&CO->trace[i], *CO_time.epochTimeOffsetMs)`,
- in some user loops, fixed numbers like `CO_NO_TRACE` or `CO_NO_TPDO` should be replaced with appropriate counts, like `CO->traceCount` or `CO->tpdoCount`.

Please note that the changes for SRDO and trace were _not_ compile-tested, however they should work (;

### DIFF
--- a/301/CO_SDOserver.h
+++ b/301/CO_SDOserver.h
@@ -422,16 +422,16 @@ typedef enum{
  * can also be used to access the OD by index like this.
  *
  * \code{.c}
- * index = CO_OD_find(CO->SDO[0], OD_H1001_ERR_REG);
+ * index = CO_OD_find(&CO->SDO[0], OD_H1001_ERR_REG);
  * if (index == 0xffff) {
  *     return;
  * }
- * length = CO_OD_getLength(CO->SDO[0], index, 1);
+ * length = CO_OD_getLength(&CO->SDO[0], index, 1);
  * if (length != sizeof(new_data)) {
  *    return;
  * }
  *
- * p = CO_OD_getDataPointer(CO->SDO[0], index, 1);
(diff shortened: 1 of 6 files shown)


FreddieChopin:
Please ignore it for now - currently I'm exploring various `objcopy` tricks to have several instances of CANopenNode code in the single project.

CANopenNode:
OK.

One info: I'm finalizing new Object Dictionary interface. I think, I found quite nice solution. This week will be published. I will ask you for a review.

Issue #205, opened 2020-07-01
SDO block download logic aspects (sequence breaks, losts etc.)


lukegluke:
@CANopenNode, I would like to discuss some aspects with SDO block download I've run into. Especially while you are working on renewed SDO server.

1. CanOpen 301 spec not well defined under which conditions the server has to send an ACK (section 7.2.4.2.10) related to sequence breaks.
2. For now CANopenNode sends ACK with seq of last correctly received segment on every (in theory) received messages with not expected sequence number
`if(seqno == (SDO->sequence + 1U))`
except case when we are waiting new subblock (or receive same segment)
`else if((seqno == SDO->sequence) || (SDO->sequence == 0U)) `
By "in theory" I mean that messages are received at high rate, while SDO process is done more rare, so in fact driver do not manage to send ACK on each message with wrong seq.
3. Spec says that
> If ACK number does not correspond with the sequence number of the last segment sent by the client during this block transfer the client shall retransmit all segments discarded by the server **within the next block transfer**.
4. SDO client in our case send data by subblocks, it can't stop in the middle (and looks like spec also expects this kind of SDO client behavior).

Keeping this in mind.

I think we shouldn't send ACK on every wrong seq if received seqno is less than expected (SDO->sequence + 1U).
If SDO client start resend subblock from very first segment (discarding for now how we got into this situation), CANopenNode periodically (form SDO_process) sends ACK back.
![изображение](https://user-images.githubusercontent.com/3624285/86271845-4ac45e00-bbd6-11ea-977f-35c3100d3b1a.png)

After client completes, it handled queued received ACK from CANopenNode SDO server, and start resend "missed" data.
CANopenNode SDO server for that time received whole subblock, ACK it [I will write ACK=127 for simplicity] and waits first segment of new block ignoring anything else.
If client SDO doesn't squash received ACKs and simply handles them just one by one, pulling them from received queue - it will restart subblock transfers quite several times for each previously received ACK (starting 107 seq in pictured case). In theory as many times as it saves segments before, but in fact much fewer.

Another issue. What if ACK of last segment of subblock gets lost?
CANopenNode SDO server will move internal sequence to 0, expecting 1 segment of new subblock. But client SDO missed ACK and is still waiting ACK. After timeout CANopenNode will send ACKseq=0, it force client to retransmit whole **previous** subblock, but CANopenNode will write it as a new one - data get corrupted.
[...]

PR #210, opened 2020-07-13
SDO: fix broken block download

lukegluke:
Fix wrong double calling CO_SDO_receive_done for last segment in subblock/block.

This is critical fix for PR #174
issue #39

Before this patch SDO block download was not working if there is no sequence breaks or timeouts (if SDO buffer not gets empty while receiving block). If data size is bigger or equal to blocksize, on receiving last segment, bug caused SDO to process wrong message buffer from FIFO while waiting block end sequence that leads to SDO abort 0504 0001h (client/server command specifier not valid or unknown).

It was caused by wrong double calling CO_SDO_receive_done on block end (one call in full buffer check and one in block end check) that broke FIFO receive pointer. Workaround was to set CO_SDO_BUFFER_SIZE greater than 127*7 (889 bytes) or work with CANopen master that always brake sequence on downloading :)

This fix preserve "old logic" we discuss in #205 issue.

### DIFF
--- a/stack/CO_SDO.c
+++ b/stack/CO_SDO.c
@@ -219,24 +219,26 @@ static void CO_SDO_receive(void *object, const CO_CANrxMsg_t *msg){
             /* check correct sequence number. */
             if(seqno == (SDO->sequence + 1U)) {
                 /* sequence is correct */
-                uint8_t i;
 
-                SDO->sequence++;
+                /* check if buffer can store whole message just in case */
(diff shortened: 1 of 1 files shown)


lukegluke:
It must be merged in old 1.3-master, but @CANopenNode it is up to you what to do with master, especially if you refactoring SDO. But take into account that current SDO block can't work at all for data longer than blksize.

CANopenNode:
I fixed the same also in master.

lukegluke:
Ok, thanks!

PR #212, opened 2020-07-21
fix for missing int cast which cause compilation warning/error when c…

Norcio:
…type,__cplusplus and esp32 is used

error: array subscript has type 'char' [-Werror=char-subscripts]


CANopenNode:
Casting to int in isdigit() and similar is more correct. Thank you.

PR #215, opened 2020-08-01
Defined ASCII Gateway Line Termination (Newline) configuration option

heliochronix:
This basically allows an application employing the 309 ASCII Gateway to define the default line termination (newline) sequence for things like help text and other returned strings.

As it was before, the line terminations were simply "\n". This works great for a Linux system, but doesn't work very well for serial connected MCUs that expect a "\r\n" sequence. By providing this configuration option in the driver target header file, the user can control the line termination sequence as appropriate for their needs.

### DIFF
--- a/309/CO_gateway_ascii.c
+++ b/309/CO_gateway_ascii.c
@@ -142,72 +142,72 @@ void CO_GTWA_log_print(CO_GTWA_t* gtwa, const char *message) {
 #if (CO_CONFIG_GTW) & CO_CONFIG_GTW_ASCII_PRINT_HELP
 /* help strings */
 static const char CO_GTWA_helpString[] =
-"\nCommand strings start with '\"[\"<sequence>\"]\"' followed by:\n" \
-"[[<net>] <node>] r[ead] <index> <subindex> [<datatype>]        # SDO upload.\n" \
-"[[<net>] <node>] w[rite] <index> <subindex> <datatype> <value> # SDO download.\n" \
-"\n" \
-"[[<net>] <node>] start                   # NMT Start node.\n" \
-"[[<net>] <node>] stop                    # NMT Stop node.\n" \
-"[[<net>] <node>] preop[erational]        # NMT Set node to pre-operational.\n" \
-"[[<net>] <node>] reset node              # NMT Reset node.\n" \
-"[[<net>] <node>] reset comm[unication]   # NMT Reset communication.\n" \
-"\n" \
-"[<net>] set network <value>              # Set default net.\n" \
-"[<net>] set node <value>                 # Set default node.\n" \
-"[<net>] set sdo_timeout <value>          # Configure SDO time-out.\n" \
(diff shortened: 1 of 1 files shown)


CANopenNode:
Are you actually running gateway on serial connected MCU? Is it working? A have actually written gateway with that in mind, but I didn't test it yet.

Yes, I used only "\n" to terminate lines in help texts, but I was not sure about that. Actually the standard requires using "\r\n" to terminate the command (which does not make problems on Linux). But I wanted to distinguish end-of-command and newlines between help texts. And maybe I just made a mess with that.
However, we must never change "\r\n" command terminators with "\n" only. Only "\n"-s inside help strings may be configurable.

heliochronix:
Yes, I've been testing the ASCII Gateway on a serial MCU (mostly for
testing on a satellite bus). I was having issues with the newlines in help
text and strings printed out for the gateway, which prompted the pull
request. I think I didn't touch the command input part, as I knew that had
to have the \r\n termination (and I left the help text regarding that part
the same). This is only for the returned string buffers, though I'd
appreciate if you checked my work in that regard.

Thanks!

heliochronix:
Just a quick follow up on this. I realized I never answered your question on whether it is working. Generally speaking, yes it is working! As far as non-help text parts are concerned, we've been able to use the gateway pretty well. There are some other issues we are having, such as a lack of a synchronous write/read, that are kinda messing with the serial interface, but besides that it's working pretty well.

I might open an issue regarding the synchronous thing, just to track it. Basically, as it stands, a simple shell built on a serial interface gets a little weird with the asynchronous callback implementation of responses. For example, a command issued via the shell will immediately return and print the shell prompt again, but this usually occurs before or in the middle of the response callback. One solution I tried was suspending the shell thread until the callback determined there was no more data to be read back, but I couldn't find a good way of determining this condition.

Overall though, I think the gateway is working pretty decently! The response newlines, and the lack of a good indicator to block on or just a straight blocking function are the only issues I've experienced so far.

CANopenNode:
It is nice to hear, it is working pretty decently.
Sorry for my late answer. You have to fix your PR. Standard requires using "\r\n" to terminate the command and response and this must not be configurable. You can make configurable only newlines "\n" inside help texts.

heliochronix:
Oh, my mistake. I didn't realize the response was also required to be `\r\n`. If that is the case, should the help text just be standardized on that as well? Is there any harm in setting that as the line delimiting sequence for everything in the ASCII Gateway?

heliochronix:
Fixed the gateway response strings for now.

It seems to me that making the help text the same as every other response string returned from the gateway would be best. It would keep things consistent. It might be a bad idea to have part of it configurable but not the other part.

CANopenNode:
I also think it would be best to just make everything the same - "\r\n" without configuration option. Can you change that.

heliochronix:
Done!

CANopenNode:
Thanks

PR #217, opened 2020-08-06
Redefine CO_SDO_state_t numbers to allow parsing of upper nibble as flags

heliochronix:
This is a subtle change to the `CO_SDO_state_t` enumeration that allows the upper nibble of the byte to be parsed as flags indicating whether the state is for an upload or download, and if it is a block transfer or not.

The idea behind this change is that a generic SDO client thread can wake up via the `CO_SDOclient_initCallbackPre()` defined callback function and check the `SDO_C->state` value to determine whether an upload or download is in progress (and optionally if it is a block type transfer). This allows the appropriate processing functions to be called without implementing a series of global variables for each SDO client in existence to externally track the state of a client thread.

Effectively this allows multiple SDO client threads to operate independently to process any number of SDO client objects, because it keeps the state internal to the client object.


heliochronix:
I just realized that "local transfer" would be ambiguous in this case, as there is no way to tell if the local transfer is a download or upload.

Hmm.. Perhaps I need to look at this a little more...

PR #227, opened 2020-09-22
fix compile error when TIME is not enabled

wilkinsw:
This fixes a compile error due to a missing guard on the CO_TIME_t component of CO_t. It also changes the rest of the TIME guards to use CO_CONFIG_TIME_ENABLE instead of CO_NO_TIME.


CANopenNode:
CO_config.h is a little future oriented.
I plan to change CANopen.c in a way you propose with next version (new OD). Also CO_time.c will be updated then.

PR #229, opened 2020-09-24
CO_HBconsumer_process - skipped monitored nodes, useless cycles

oto313:
When current node is CO_HBconsumer_UNCONFIGURED then other nodes are skipped. You can increment pointer to node or break the cycle.


CANopenNode:
Thank you for this. It is another bug in heartbeat consumer. I will fix this also in v1.3.

PR #231, opened 2020-09-30
NMT: On startup, set state to operational when master

jdascenzio:
This problem was introduced by 06a0493c787e6232a04d1e24376854d8339f1e80
Since this commit, the master don't start to operational state

### DIFF
--- a/stack/CO_NMT_Heartbeat.c
+++ b/stack/CO_NMT_Heartbeat.c
@@ -226,8 +226,9 @@ CO_NMT_reset_cmd_t CO_NMT_process(
             if(HBtime > NMT->firstHBTime) NMT->HBproducerTimer = HBtime - NMT->firstHBTime;
             else                          NMT->HBproducerTimer = 0;
             
-            /* NMT slave self starting */
-            if (NMTstartup == 0x00000008U) NMT->operatingState = CO_NMT_OPERATIONAL;
+            /* NMT slave self starting or master */
+            if ((NMTstartup == 0x00000008U) || (NMTstartup & 0x00000001U))
+                                           NMT->operatingState = CO_NMT_OPERATIONAL;
             else                           NMT->operatingState = CO_NMT_PRE_OPERATIONAL;
         }
     }


CANopenNode:
NMTstartup (Object 0x1F80) is specified by CiA302-2, which describes quite complex CANopen manager (master), bit0.

In CANopenNode is available only a simple NMT master, where application can start remote nodes, including itself, bit1.

Self start of slave is controlled by bit3.

I don't like to implement bit0 for self-start, as it is not according to standard. I think so.

Maybe self-start can be implemented somewhere in the application, if needed.
In the next version I will probably replace "NMTstartup" argument to CO_NMT_process() function with simple bool value or something.

jdascenzio:
I understand your arguments but for the moment, in this branch, the master always remains in the pre-operational state without, it seems to me, no real solution to switch it to operational mode except by directly setting the variable CO-> NMT-> operatingState. Another solution might be to add a function such as: CO_NMT_setInternalState....

CANopenNode:
Yes those are for now the options.
I don't like to change it in current version, because existing devices may start behave differently.

But for upcoming release with new OD such things will be no problem. I'm thinking about changing NMTstartup to custom argument: a bitfield, which will define startup state on the beginning and will also control state in runtime (automatic enter to operational, if errors are gone). Someone could still use 1F80 object, but in own way.

CANopenNode:
But we can add CO_NMT_setInternalState(), which is just a shortcut to NMT-> operatingState = ...

jdascenzio:
> But we can add CO_NMT_setInternalState(), which is just a shortcut to NMT-> operatingState = ...

I create a PR for that:  #232

CANopenNode:
Thanks

martinwag:
Sorry, just going trough this now. You can set your master application to operational state by calling CO_sendNMTcommand() with your own node ID. So I think we don't need another way to do that.

jdascenzio:
You are quite right! So the commit 09f1083ae661fdbc93731840be10fe45aeb678df is not necessary...

CANopenNode:
CO_sendNMTcommand() also sends a CAN NMT message and it is part of optional NMT master. 
But CO_NMT_setInternalState() just simply changes the state and it is in pair with CO_NMT_getInternalState()

CANopenNode:
I think, it is OK now. CO_NMT_setInternalState() is also available in next release.

PR #237, opened 2020-10-08
Make "remote NMT changed" callback of HB consumer node-specific

FreddieChopin:
* This way "remote NMT changed" callback is much more similar to other "multi" callbacks of HB consumer, and it was node-specific anyway, as you have to first configure the monitoring node to have that callback executed.

* Merge CO_CONFIG_HB_CONS_CALLBACK_CHANGE with ..._HB_CONS_CALLBACK_MULTI
These settings are now very similar, so there's no need to have two separate config options.

* Implement a better solution for #229
Instead of playing with incrementing pointers in various places it is better to just set it once at the start of loop.

These changes break the API.

### DIFF
--- a/301/CO_HBconsumer.c
+++ b/301/CO_HBconsumer.c
@@ -125,9 +125,6 @@ CO_ReturnError_t CO_HBconsumer_init(
     HBcons->NMTisPreOrOperationalPrev = false;
     HBcons->CANdevRx = CANdevRx;
     HBcons->CANdevRxIdxStart = CANdevRxIdxStart;
-#if (CO_CONFIG_HB_CONS) & CO_CONFIG_HB_CONS_CALLBACK_CHANGE
-    HBcons->pFunctSignalNmtChanged = NULL;
-#endif
 
     for(i=0; i<HBcons->numberOfMonitoredNodes; i++) {
         uint8_t nodeId = (HBcons->HBconsTime[i] >> 16U) & 0xFFU;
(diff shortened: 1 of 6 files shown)


CANopenNode:
Hi Freddy,

I would like to revert your commit "Make "remote NMT changed" callback of HB consumer node-specific".

```CO_CONFIG_HB_CONS_CALLBACK_CHANGE``` was intended to be simplified alternative to ```CO_CONFIG_HB_CONS_CALLBACK_MULTI```. I would like to keep them separated.

FreddieChopin:
Please let me know whether you want to revert all of it or just the commit which merged the two config options. If you want to revert it all, I propose an alternative - make the options mutually exclusive - you can either have one common callback via CO_CONFIG_HB_CONS_CALLBACK_CHANGE or you can have all the separate callbacks (including the one for "change") via CO_CONFIG_HB_CONS_CALLBACK_MULTI.

However I don't think that the previous behaviour was any easier to use... If this callback would work WITHOUT configuring the HB consumer, then sure, but you have to configure it anyway. For a truly object-oriented application the previous behaviour (one common callback for all consumers) is very awkward to use and I really don't see anything significantly harder in using the new variant. Plus there's a bonus of having one configuration option less.

CANopenNode:
Thank you for your feedback. Also all this configuration options are my invention and I'm not 100% sure, what they mean for others. So if you have any further critics or suggestions, please let me know.

Otherwise, I agree with your alternative proposal. I would like to have simple alternative just for logging the changes.

FreddieChopin:
One question then. Previous signature of the "change" callback was `void(uint8_t nodeId, CO_NMT_internalState_t NMTstate, void *object)`. This pull request changed it to `void(uint8_t nodeId, uint8_t idx, CO_NMT_internalState_t NMTstate, void *object)` (it added the index). Should the simpler callback have the simpler (previous) signature, or the new one (to be consistent)?

CANopenNode:
No problem, can have the new signature.

PR #238, opened 2020-10-08
If CO_NO_TIME is 0, then CO object does not have TIME field

FreddieChopin:


CANopenNode:
Already fixed in #234

PR #241, opened 2020-10-24
Restore CO_CONFIG_HB_CONS_CALLBACK_CHANGE option

FreddieChopin:
The user may decide whether he/she wants one common callback for all
monitored nodes (CO_CONFIG_HB_CONS_CALLBACK_CHANGE) or maybe separate
callbacks configured individually for each monitored mode
(CO_CONFIG_HB_CONS_CALLBACK_MULTI). These options cannot be set
simultaneously.

This change is related to the discussion in #237

Also fix a warning about shadowed local variables in CO_LSSslave.c

### DIFF
--- a/301/CO_HBconsumer.c
+++ b/301/CO_HBconsumer.c
@@ -31,6 +31,12 @@
 
 #if (CO_CONFIG_HB_CONS) & CO_CONFIG_HB_CONS_ENABLE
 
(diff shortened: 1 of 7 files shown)


CANopenNode:
Thank you for this. Those two changes were also added to latest newOD branch.

PR #249, opened 2020-12-02
fixed CO_fifo compiler warnings

geoffrey-vl:


CANopenNode:
Applied also to newOD branch

PR #250, opened 2020-12-02
fix: XC16 compiler error

JuPrgn:
error: 'for' loop initial declarations are only allowed in C99 mode

### DIFF
--- a/301/CO_HBconsumer.c
+++ b/301/CO_HBconsumer.c
@@ -352,7 +352,8 @@ void CO_HBconsumer_process(
     uint8_t allMonitoredOperationalCurrent = CO_NMT_OPERATIONAL;
 
     if (NMTisPreOrOperational && HBcons->NMTisPreOrOperationalPrev) {
-        for (uint8_t i=0; i<HBcons->numberOfMonitoredNodes; i++) {
+        uint8_t i = 0;
+        for (i=0; i<HBcons->numberOfMonitoredNodes; i++) {
             uint32_t timeDifference_us_copy = timeDifference_us;
             CO_HBconsNode_t * const monitoredNode = &HBcons->monitoredNodes[i];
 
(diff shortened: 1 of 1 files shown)


CANopenNode:
I didn't test Microchip compiler yet.
Can you try, if it is possible to set C99 mode in the compiler?

JuPrgn:
Hi Janez,

I didn't know how to enable this mode on the XC16 compiler, a note was attached to the error :
`note: use option -std=c99 or -std=gnu99 to compile your code`
But using -std=c99 option gave errors with "asm".
The working method to enable this is to add `-std=gnu99`  in Properties > XC16 (Global Options) > Additional options.
Please note that this is not documented in the XC16 User's guide (it does not appear to be an official option) and may change other things .

I usually prefer to keep default compiler options but I guess you use a lot of initialisation inside loops on new stack ?
We could document this on CANopenPIC doc if you prefer to go this way.

CANopenNode:
Hi,

We had some discussion about C99 support: https://github.com/CANopenNode/CANopenNode/issues/162#issuecomment-658594880

If microchip enables C99, I would prefer to use its option and document on CANopenPIC doc.

I'm currently working on Linux and after I finish here, I will make a complete updates in all PICs too. I have plans to make an example for https://store.digilentinc.com/max32-arduino-programmable-pic32-microcontroller-board/. Do you know any similar board for 16-bit PICs (besides explorer16 board)?

JuPrgn:
Ok this is true that C99 would be great but Microchip are always a bit late (20 years really ?!).
I admit that doing the ephemeral declarations in loops is really cool and it improves the readability of the code.

I made a PR to add some declarations for your PIC drivers but I did not had many changes to switch from v1.3 and that's great ! 

There are a lot of new generic boards (PIC and others) on the web for different kind of users but I cannot help you much on this area because I am always working on custom embedded boards. 
I'll keep an eye on these and let you know if I find anything of interest.

JuPrgn:
I updated my PR https://github.com/CANopenNode/CANopenPIC/pull/7 to add C99 configuration.

PR #251, opened 2020-12-03
MNT: Add a function to set internal state

geoffrey-vl:
Same as https://github.com/CANopenNode/CANopenNode/pull/232, but now in master branch.


CANopenNode:
Those two functions were already in newOD branch and were today copied from there also to master branch. Thank you anyway.

PR #253, opened 2020-12-10
Fix Conformance Test Tool error on test SDO 2.9

JuPrgn:
Reset errors history instead of returning CO_SDO_AB_NO_DATA

### DIFF
--- a/301/CO_Emergency.c
+++ b/301/CO_Emergency.c
@@ -88,9 +88,6 @@ static CO_SDO_abortCode_t CO_ODF_1003(CO_ODF_arg_t *ODF_arg){
         if(ODF_arg->subIndex == 0U){
             ODF_arg->data[0] = noOfErrors;
         }
-        else if(ODF_arg->subIndex > noOfErrors){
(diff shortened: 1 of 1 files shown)


CANopenNode:
I'm not sure about this. I didn't run Conformance Test Tool, but standard clearly says:

> NOTE: If no error is present the value of sub-index 00 h is 00 h and a read access to sub-index 01 h is responded with an SDO abort message (abort code: 0800 0024 h or 0800 0000 h ).

Maybe the problem is in EDS file, where object 0x1003, 00 should have value of 0.

JuPrgn:
Thank you for your answer I will look further into this next week and check the conformance test plan to clarify this.

JuPrgn:
You are correct SDO 2.9 attempt to write a value to an RW entry and since 1003 only accepts the value "0" when writing this test fails. The good practice is therefore to modify the EDS file, I was using the number of sub indexes for display on my HMI but I will have to find another way better suited to this specific object. Thanks

PR #254, opened 2020-12-15
Ci build

geoffrey-vl:
We can now have free server builds using github's CI pipeline.
The current CI build compiles the linux stack on each push and pull request.


CANopenNode:
Thank you

PR #265, opened 2021-02-16
BUGFIX: updating nmt internal state doesn't trigger callback function

geoffrey-vl:
Noticed that when we set the nmt internal state directly that the attached callbacks are not triggered.
Easiest way to fix is as I did: trigger the callback from the setter function.

Let me know if the solution is not satisfying.

### DIFF
--- a/301/CO_NMT_Heartbeat.c
+++ b/301/CO_NMT_Heartbeat.c
@@ -328,6 +328,25 @@ CO_NMT_reset_cmd_t CO_NMT_process(
     return (CO_NMT_reset_cmd_t) NMT->resetCommand;
 }
 
+/******************************************************************************/
+void CO_NMT_setInternalState(CO_NMT_t *NMT,
+                                           CO_NMT_internalState_t state)
+{
+    if (NMT == NULL) 
+        return;
+
+    if (NMT->operatingState == state)
+        return;
+
+    NMT->operatingState = state;
+
+#if (CO_CONFIG_NMT) & CO_CONFIG_NMT_CALLBACK_CHANGE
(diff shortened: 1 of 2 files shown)


CANopenNode:
I think, the solution is OK.

The same problem was also in newOD. I solved it there too, but with different approach: `CO_NMT_setInternalState()` was replaced by `CO_NMT_sendInternalCommand()`. Function sets internal variable `NMT->internalCommand`, which is verified and processed by `CO_NMT_process()`

geoffrey-vl:
> I think, the solution is OK.
> 
> The same problem was also in newOD. I solved it there too, but with different approach: `CO_NMT_setInternalState()` was replaced by `CO_NMT_sendInternalCommand()`. Function sets internal variable `NMT->internalCommand`, which is verified and processed by `CO_NMT_process()`

Well I noticed today that using my approach is not the best solution. For example the basic application loop which mostly look like:

```
while(resetCmd != CO_RESET_APP && resetCmd != CO_RESET_QUIT && gCO_endProgram == 0)  {
...
    while(resetCmd == CO_RESET_NOT && gCO_endProgram == 0) {
        ....
        CO_epoll_processMain(&_epMain, CO, &resetCmd);
        ...
    ...
    }
}
```
... will not pickup the changed NMT state. Hence, when you're trying to call a communication reset, the NMT state changes but it doesn't trigger a real comms reset.

@CANopenNode as I understand in V4 this is fixed?

CANopenNode:
Sorry for my very late answer. I was just too busy with other than CANopen.

> @CANopenNode as I understand in V4 this is fixed?

Yes, with approach in v4 everything should work correctly.

PR #266, opened 2021-02-18
Fix build in NewOD for certain combinations of options

dergentk:
Some structure members are only present when explicitly enabled by options.
Consequently, these members can only be used when the corresponding option is enabled.
I found some instances where this check in the code was missing and thus resulted in compile errors


CANopenNode:
Thank you for this.
CO_Emergency was more tricky, CANerrorStatus should always be checked, so CANdevTx must not be optional. I fixed it.

PR #277, opened 2021-03-19
Warning fixes

### LINKED ISSUE #275: Null pointer dereference
I found some null pointer dereference on New OD I will propose a PR to fix some of them.

On this one there is a dereference before checking for `NULL` : 

https://github.com/CANopenNode/CANopenNode/blob/66d0e14ef4ac631894c5b4adc0441ef4406c2eb9/301/CO_SDOserver.c#L686-L691

Do you prefer to return at the beginning of the function checking the `NULL` parameters or do you want to keep only one exit point ?

JuPrgn:
Fix #275 

I separated the different type of fix then you can choose if you do not want to apply something.

I could not run clang-format as this change a lot of files according to `ColumnLimit:` maybe you want to adjust this value on `.clang-format` ?

### DIFF
--- a/301/CO_ODinterface.c
+++ b/301/CO_ODinterface.c
@@ -145,7 +145,6 @@ OD_entry_t *OD_find(OD_t *od, uint16_t index) {
         return NULL;
     }
 
-    uint16_t cur;
     uint16_t min = 0;
     uint16_t max = od->size - 1;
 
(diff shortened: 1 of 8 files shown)


CANopenNode (approved):

CANopenNode (approved):

CANopenNode:
Thank you for the fixes.

I didn't strictly follow clang-format as I rather format manually myself. Don't like automatic formatter much.
I was quite strict with column limit = 80 characters in all new code. But if code was not changed, I did not reformat the file.

JuPrgn:
You are welcome.

Ok, I understand your point, sometimes clang-format modifies things that we would prefer to keep unchanged but it is very useful to keep consistency when many people contribute. It is not a priority for the moment but I will try to find a config close to the current format.

PR #286, opened 2021-03-24
Fix HBcons entries count

JuPrgn:
HB Consumer sub index 0 is `Highest sub-index supported` and should not be counted on number of HB consumers.
Without this, `CO_CANopenInit()` fails with error code -12 : `CO_ERROR_OD_PARAMETERS`


CANopenNode:
You are right. There is a confusion with `subEntriesCount`, it is (was) documented wrong.

PR #295, opened 2021-04-27
fixed some warnings & potential bugs

gotocoffee1:
Hello, I fixed some warnings, that occured on my setup. A good bunch are just just typecasts, mostly because 
`sizeof(OD_size_t) != sizeof(size_t)` on 64 bit. I added the casts also in 301/CO_SDOclient.c but I am not sure if it is better to switch the function signatures instead (would also be consistent with the server part). 

bugs:

* 301/CO_SDOclient.c L 415 set invalid after return (noreachable code)
* 301/CO_SDOserver.c L 342 could lead to NULL access
* 301/CO_NMT_Heartbeat.hL 318 does not compile without cast when included in cpp file


CANopenNode:
Thank you.
I will also add null checking into OD_getIndex() function. (related to "301/CO_SDOserver.c L 342 could lead to NULL access").

PR #301, opened 2021-05-21
remove warnings in CO_LSS (backport from v4)

geoffrey-vl:
Some part of https://github.com/CANopenNode/CANopenNode/commit/267a2dafb29124b814215c5046922075ea03b48f about emitting warnings is now backported to v1.3

### DIFF
--- a/stack/CO_LSS.h
+++ b/stack/CO_LSS.h
@@ -144,7 +144,7 @@ typedef enum {
     CO_LSS_FASTSCAN_CONFIRM = 0x80U  /**< All LSS slaves waiting for scan respond and previous scan is reset */
 } CO_LSS_fastscan_bitcheck;
 
-#define CO_LSS_FASTSCAN_BITCHECK_VALID(bit) ((bit>=CO_LSS_FASTSCAN_BIT0 && bit<=CO_LSS_FASTSCAN_BIT31) || bit==CO_LSS_FASTSCAN_CONFIRM)
+#define CO_LSS_FASTSCAN_BITCHECK_VALID(bit) (bit<=CO_LSS_FASTSCAN_BIT31 || bit==CO_LSS_FASTSCAN_CONFIRM)
 
 /**
  * Fastscan LSSsub, LSSnext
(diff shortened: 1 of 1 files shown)


martinwag:
I don't know about v4.0, but for the v1.3 code this removes a part of range checks, which, in my opinion makes the code wrong. 
The range check macros are in the public part of the code and can also be used with signed integers.

ps: I'm the one that has written the LSS code...

geoffrey-vl:
Well the compiler is warning us: `comparision is always true: index >= CO_LSS_BIT_TIMING_1000`.
I noticed it was patched in V4 several months ago hence why I thought was all verified and OK.

martinwag:
I'm not actively verifying code for v4.0, so I didn't see that one. From what I've seen in a short look, LSS is mostly(?) unchanged. The change will also not break the stack code, as the comparision macro in this case will evaluate to "uint >= 0". I'm fully aware of that.

Removing the lower bound check in this case has the following problems, in my opinion:
- Macros don't have data types. The problematic macros are in the header file, so the stack user can use them in own code anywhere with any data type, including signed ones. Macros should always be save to use, so requires check for negative value.
- My intention as a programmer was to do a range check, not an upper limit check. From the macro itselfe one can't even see that one of the values is zero. Removing one of the range limits means it should be at least part of the code comment, which I think is a less good option than being part of the code.

The solution would be to use inlines instead of macros. However, last time I tried to use inlines they required a different declaration for different compilers...

geoffrey-vl:
I agree with your opinion here, I would also insist on having the right solution fixed in all branches all together then.

I'm not sure about your solution, as you said using `inline` may be problematic. In the end I haven't found any inline functions for now in the stack, except for socketcan drivers. Would you agree with moving the MACROs to the source files as a solution, so that it is assured no one can reuse that code outside of its context? It's a quick fix and doesn't affect code readability.

Also I noticed that the macro's aren't used that much, so another solution may be to remove the macros altogether and just copy the code a few times which assures no one can misuse them but creates duplicate code and maybe less clean code?

martinwag:
We can also follow the GCC documentation [1](https://gcc.gnu.org/onlinedocs/gcc/Warning-Options.html)

> -Wall
> This enables all the warnings about constructions that some users consider questionable, and that are easy to avoid (or modify to prevent the warning), **even in conjunction with macros**.

and use Wall instead of Wextra.  In my opinion the stuff that Wextra checks should be checked by static code analysis, not by the compiler. This has the advantage of having the macro available for parsing.
As we don't have CI / static code analysis for the stack, I would agree with you to move the macros inside the corresponding c files. I'd prefer to keep the macros for code readability reasons, even if a macro is only used once.

CANopenNode:
Yes, Martin is the author of all LSS. Code is mostly unchanged. LSS slave is now part of all example devices, node-id and bit-rate storage is implemented by default, tutorial is updated and there is more descriptive output from scan-all-nodes.

Those compiler warnings were a little annoying for me too. I will check that next week.

CANopenNode:
It was a long "week".
I made in v4 as you suggested, see commit above. I think, we can backport the same also to v2 and v1.3.

CANopenNode:
This is an old PR, I'm closing it. If still relevant, please update, test, and re-open this.

PR #305, opened 2021-05-28
Fix compile errors

cfr-mir:
Defining CO_MULTIPLE_OD or removing CO_CONFIG_PDO_IO_ACCESS.
Generates compile errors.

### DIFF
--- a/301/CO_PDO.c
+++ b/301/CO_PDO.c
@@ -71,7 +71,6 @@ static ODR_t OD_read_dummy(OD_stream_t *stream, void *buf,
     return ODR_OK;
 }
 
-
 /*
  * Find mapped variable in Object Dictionary and configure entry in RPDO or TPDO
  *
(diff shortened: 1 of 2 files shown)


CANopenNode:
Hello,
I'm very late with my response, I just had no time for CANopenNode in the last month.

I think, second commit is not necessary. If one disables CO_CONFIG_PDO_OD_IO_ACCESS, then CO_CONFIG_GLOBAL_FLAG_OD_DYNAMIC should be disabled also. There is also an error message for this: `#error Dynamic PDO mapping is not possible without CO_CONFIG_PDO_OD_IO_ACCESS`

CANopenNode:
Configuration with CO_MULTIPLE_OD is now fixed and tested. I don't like to add `CO_CONFIG_PDO_OD_IO_ACCESS` for the reasons described above. I made some other changes, so this PR is no more relevant.

PR #306, opened 2021-05-28
Only initialize time if CO_GET_CNT(TIME)==1

cfr-mir:
If NO time service is defined in the OD and the TIME service feature is not enabled.
A rx buffer is allocated.

### DIFF
--- a/CANopen.c
+++ b/CANopen.c
@@ -550,7 +550,9 @@ CO_t *CO_new(CO_config_t *config, uint32_t *heapMemoryUsed) {
 #endif
         co->RX_IDX_EM_CONS = idxRx; idxRx += RX_CNT_EM_CONS;
 #if (CO_CONFIG_TIME) & CO_CONFIG_TIME_ENABLE
-        co->RX_IDX_TIME = idxRx; idxRx += RX_CNT_TIME;
+        if (CO_GET_CNT(TIME) == 1) {
+        	co->RX_IDX_TIME = idxRx; idxRx += RX_CNT_TIME;
+        }
 #endif
 #if (CO_CONFIG_GFC) & CO_CONFIG_GFC_ENABLE
         co->RX_IDX_GFC = idxRx; idxRx += RX_CNT_GFC;


CANopenNode:
We have to discuss about this commit too. You can't do that, because CO_TIME_init() function uses idxRx so rx buffer could get overlapped with other CANopen object.

Configuration may seems complex. There are two sources of configuration:
1. primary: CO_CONFIG... macros, which enables/disables parts of code
2. secondary: object dictionary, which may or may not have specific objects.

I tried to make stack independent from object dictionary, so there is not much automation. 
Optimization of the stack must be done with CO_CONFIG... macros (in CO_driver_target.h or CO_driver_custom.h file).
Object dictionary must then match the configuration somehow. If there are missing required OD entries, then stack may compile but will not run from the first time.

CANopenNode:
Sorry, I missed:
```
#if (CO_CONFIG_TIME) & CO_CONFIG_TIME_ENABLE
    if (CO_GET_CNT(TIME) == 1) {
        err = CO_TIME_init(co->TIME,
...
```
I will check again, if we can optimize rx buffer usage.

CANopenNode:
Actually this commit is not necessary, because
RX_CNT_TIME gets value of 1 only `if (CO_GET_CNT(TIME) == 1)`, otherwise it has value 0 and additional rx buffer is not allocated.

(But I must say, I'm not 100% sure if it works correctly, I didn't use MULTI_OD much.)

cfr-mir:
It didn't work for me, with MULTI_OD defined, hence the PR.
Time service was not  initialized, buf the buffer for it was allocated.
I may have been the cause of this myself.

I will look into it, but as you, I get occupied elsewhere.

________________________________
Fra: Janez ***@***.***>
Sendt: 22. juni 2021 10:29
Til: CANopenNode/CANopenNode ***@***.***>
Cc: Carsten Frank ***@***.***>; Author ***@***.***>
Emne: Re: [CANopenNode/CANopenNode] Only initialize time if CO_GET_CNT(TIME)==1 (#306)

CAUTION - EXTERNAL EMAIL: Do not open attachments or links unless you recognize the sender and the content is safe.



Sorry, I missed:

#if (CO_CONFIG_TIME) & CO_CONFIG_TIME_ENABLE
    if (CO_GET_CNT(TIME) == 1) {
        err = CO_TIME_init(co->TIME,
...


I will check again, if we can optimize rx buffer usage.

CANopenNode:
Actually this PR is not necessary, rx buffer is not allocated, if CO_GET_CNT(TIME) == 0. I was working with CO_MULTIPLE_OD in the last days and have included all your ideas and some mine, see next PR.

PR #307, opened 2021-05-28
Dynamic monitored nodes

cfr-mir:
a) Fixed the ability to have more than one monitored node.
b) Added a feature to allocate dynamic memory for the monitoredNodes array.
This is an advantage if you have less than CO_CONFIG_HB_CONS_SIZE nodes you want to monitor.
If you want to monitor more than CO_CONFIG_HB_CONS_SIZE nodes. If you have multiple OD's. and
you monitor a large amount of node in one OD and few in the other. You would waste a lot of memory, if you just raise CO_CONFIG_HB_CONS_SIZE.

### DIFF
--- a/301/CO_HBconsumer.c
+++ b/301/CO_HBconsumer.c
@@ -117,16 +117,22 @@ static ODR_t OD_write_1016(OD_stream_t *stream, const void *buf,
 
 /******************************************************************************/
 CO_ReturnError_t CO_HBconsumer_init(CO_HBconsumer_t *HBcons,
-                                    CO_EM_t *em,
+									CO_EM_t *em,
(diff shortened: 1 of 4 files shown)


CANopenNode:
I did some work related to this PR in the last days. I followed your idea and optimized memory allocation for heartbeat consumer (object 1016) and also for error history (object 1003), which is similar. For this to work, also OD editor has been updated, see #314.

However, I had to do that my way, because solution is now more generic,I get rid of two configuration settings, etc. I hope, you will like the solution.

I think HB consumer is now configurable enough, except I can't simply include allocating more than one heartbeat consumer inside CANopen.c file. If you need additional, simply allocate own in your application. (Note that one HB consumer is connected with one 1016 OD object (or similar) and each HB consumer can have configurable size of monitored nodes now.)

Thank you for PRs and for testing.

PR #312, opened 2021-06-24
PDO_initMapping() fix + CO_CANopenInitPDO() call in example + warning fix

FreddieChopin:


CANopenNode:
Thank you for your PR. I just missed the CO_CANopenInitPDO() in example.

In commit "Fix error in PDO_initMapping()" I would like a different approach. The problem is, that all mapping is configured in initialization phase. For example, if there are four mapping entries configured and mappedObjectsCount is set to 3, three mappings will be after the initialization. But if mappedObjectsCount will be set to 4 in run time (which is possible and sometimes useful), fourth entry will be unconfigured, which is an error. That is the reason of using CO_PDO_MAX_MAPPED_ENTRIES instead of mappedObjectsCount.

Solution may be just adding additional condition after `OD_get_u32()`:
```
if (odRet == ODR_SUB_NOT_EXIST) {
    continue;
}
```

Also please replace two `\t` in `Add CO_CANopenInitPDO() call to example's main()`

FreddieChopin:
> Also please replace two \t in Add CO_CANopenInitPDO() call to example's main()

Done.

> In commit "Fix error in PDO_initMapping()" I would like a different approach. The problem is, that all mapping is configured in initialization phase. For example, if there are four mapping entries configured and mappedObjectsCount is set to 3, three mappings will be after the initialization. But if mappedObjectsCount will be set to 4 in run time (which is possible and sometimes useful), fourth entry will be unconfigured, which is an error. That is the reason of using CO_PDO_MAX_MAPPED_ENTRIES instead of mappedObjectsCount.
> 
> Solution may be just adding additional condition after OD_get_u32():

Hmm... I may not understand some details, but for me both solutions give exactly the same end-effect. The function which I modified is called only during initialization. If I would change it the way you propose, the behaviour would be mostly identical, with the exception of not calling `OD_get_u32()`, and this particular function doesn't initialize anything or change any global state. Additionally, there is another version of `PDO_initMapping()` below, for case when `CO_CONFIG_PDO_OD_IO_ACCESS` is disabled. This additional version has the same loop condition as the code in my patch. Also the value read as `mappedObjectsCount` (subindex 0 from indexes 0x1600 and above) is read from current contents of the OD, so if someone modifies mapping at runtime, then I suppose that the value read into `mappedObjectsCount` will be different than the  initial one. But all of that doesn't really seem to matter, as this function is not called when you modify the PDO mapping at runtime via SDO.

If I'm wrong, please elaborate and I'll change it the way you propose.

CANopenNode:
Yes, there are some details. One is dynamic PDO mapping, this means changing PDO mapping in run time, without communication reset. Second PDO_initMapping() is active, where dynamic PDO mapping is not used (and also CO_CONFIG_PDO_OD_IO_ACCESS is disabled), that is why it works as you wrote.

I will try to describe problematic situation:
- at the initialization phase four PDO mapping objects are set to something and number of mapped objects are set to 2. This means, only two mapping objects matters on the startup. The third and fourth mapping objects may be configured or not, it doesn't matter on this point, as you say.
- But suppose, the third and the fourth mapping objects are prepared and valid and someone wants to expand the PDO, number of mapped objects (mappedObjectsCount) to 4. `OD_write_PDO_mapping()` is called, with `stream->subIndex == 0`. Just change mappedObjectsCount form 2 to 4, no mapping re-configuration, as it is already correct. I that case `PDOconfigMap()` is not called, as it is supposed it was called at the initialization phase already...

`PDOconfigMap()` is called at the initialization phase for each mapping object, not only for `mappedObjectsCount`. `PDOconfigMap()` is also called for individual mapping, when it is written in run-time.

FreddieChopin:
Done.

PR #320, opened 2021-07-14
Fixed bug with CANrxData buffer indexing

Maldus512:
The `CANrxData` field of the `CO_RPDO_t` struct is an array of `uint8_t` buffer with two elements if the node is SYNC-enabled; when copying the content of an RPDO to the object dictionary `CANrxData` is mistakenly indexed with an extra `[0]`, which ends up pointing to either the first or the second byte of the first buffer, resulting in erroneous RPDO transfers.


CANopenNode:
Thank you.

PR #327, opened 2021-08-13
Update CO_NMT_Heartbeat.c

Simon-Thome:
add missing typecast, fixes compiler warning "enumerated type mixed with another type"


Simon-Thome:
Alternatively, should uint8_t operatingState in struct  CO_NMT_t be changed to enum type CO_NMT_internalState_t?

CANopenNode:
I changed and pushed as in your alternative suggestion. I also changed `internalCommand` to enumerator.

PR #354, opened 2021-10-25
Remove usage of VLA in CO_storageEeprom_init().

liampwll:


CANopenNode:
Don't like to accept this PR. It reads differently from eeprom. Besides that, in most cases there is no very large array, entriesCount is not so large.

PR #355, opened 2021-10-25
Workarounds for bugs and limits in xc8.

liampwll:
These changes allow xc8 to compile CANopenNode. Perhaps not suitable to be merged but helpful for anyone that comes looking.


trojanobelix:
What version do you use? I think that might be a very old version or you are using the legacy c18 compatible mode. The newer versions >2.0  shouldn´t have problems with the original code.

PR #358, opened 2021-10-27
Create CO_process_initCallbackPre().

liampwll:
Create CO_process_initCallbackPre(). This function calls initCallbackPre functions for all functions called by CO_process() where they have been enabled. This simplifies application development where the user does not wish to call the functions in CO_process() individually but does want callbacks from those functions.


CANopenNode:
Could be useful, but CANopen.c is already complex. I prefer this driver specific.

PR #363, opened 2021-11-05
Hbchanges

cmorganBE:
Changes to let me use the CON stack to generate repeated heartbeat timeouts for devices that aren't responding to heartbeats.


trojanobelix (commented):
**The changes are not compliant with the CiA DS301 standard!**

DS301 stated that:
_Monitoring of the heartbeat producer shall start after the reception of the **first** heartbeat._

So it is not allowed that heartbeat timeouts occur on nodes that have been configured for heartbeats but for whom heartbeats have not been received.

cmorganBE:
Closing as this is against the standard.

PR #405, opened 2022-07-14
TPDO inhibitTime_us / evnetTime_us calculation fix

Blazej1994:
Added inhibitTime and eventTime casting to uint32_t


CANopenNode:
Thank you

PR #432, opened 2023-03-17
Feature/always lock od config

mlout:
Add CO_CONFIG_FLAG_ALWAYS_LOCK_OD which can be used to lock the Object Dictionary for every SDO read/write operation. This is needed when you have a RTOS with multiple tasks accessing the Object Dictionary.
Modify the code in CO_SDOserver.c and CO_SDOclient.c to check this config flag.


GiulioDallaVecchia:
It would be better to avoid any sharing of resources using an event-driven system. In that manner you don't need to use any mutex to protect the resources avoiding any problem such as starvation, priority inversion...ecc...

PR #434, opened 2023-03-29
Fix example tx bus warning checking

elotom:
I think there is copy-paste error in the example and the `rxErrors` should be replaced by `txErrors` when the tx bus warning is checked.


trojanobelix (approved):
I think the changes are correct and should be included in the code. 
Looks like a copy and paste error.

PR #466, opened 2023-06-28
Overrun not correctly reported in CO_Emergency.c

H-Grobben:
Wrong code reported during overflow.

If an overflow occurs the error code is filled with the bit value, not a error code.


trojanobelix (approved):
I think the changes are correct and should be included in the code. 
Looks like a copy and paste error.

CANopenNode:
Your update is now on master branch, thank you.
(I had to make own commit for this.)

PR #487, opened 2023-12-21
SRDO update

nicolazande:
Update SRDO module to take into account previous SDO and PDO module updates.

### DIFF
--- a/304/CO_SRDO.c
+++ b/304/CO_SRDO.c
@@ -1,844 +1,1788 @@
 /**
- * CANopen Safety Related Data Object protocol.
+ * \file CO_SRDO.c
  *
- * @file        CO_SRDO.c
- * @ingroup     CO_SRDO
- * @author      Robert Grüning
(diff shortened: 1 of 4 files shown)


nicolazande:
Last year I used your library for a project and it was great. However, I had some problems with the SRDO module. I think it was not up to date with the SDO and PDO module updates and, since I had to partially rewrite and test it, I think it could be useful also for others.
Thanks!

CANopenNode:
Thank you for you contribution. You are right, SRDO was not up to date with current CANopenNode. 

I can't merge directly into CANopenNode, because it could quite change the current formatting and so break the git history. Now I don't have much time but later I would like to carefully integrate it. For now it is here for anyone interested.

nicolazande:
OK perfect. Maybe I can remove some non-fundamental changes in the CANopen.c/h module so the SRDO related changes are clearer.

CANopenNode:
SRDO update is now on the master branch.

PR #501, opened 2024-03-11
Misra 

NLI-BrianLinari:
Make MISRA compliant for
MISRA C 2004 rule 14.10
MISRA C 2004 rule 14.9
MISRA C 2004 rule 13.1
MISRA C 2004 rule 15.3

### DIFF
--- a/301/CO_Emergency.c
+++ b/301/CO_Emergency.c
@@ -209,6 +209,7 @@ static ODR_t OD_read_1003(OD_stream_t *stream, void *buf,
         else if (index >= (em->fifoSize)) {
             return ODR_DEV_INCOMPAT;
         }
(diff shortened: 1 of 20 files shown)


martinwag:
Thanks, you've found how to create a PR :-). Those are quite many changes.

- did you check/modify all stack files or just a selection?
- is this "only" making the code MISRA rule compliant or did you do any fixes/changes?

NLI-BrianLinari:
I tried to keep similar styling as, it is different in some of the files. Do you want me to update the style guide to include the changes?

CANopenNode:
Yes, some update in the style guide would be nice. Above Misra rules should be mentioned and future changes should not break them.

CANopenNode (approved):

Issue #503, opened 2024-03-14
Porting SRDO to v4.0


temi54c1l8:

temi54c1l8:
[PortingSRDOto4.0.patch](https://github.com/CANopenNode/CANopenNode/files/14598593/PortingSRDOto4.0.patch)

CANopenNode:
Can you please make a fork and a pull request. Please note, there already is a pull request for SRDO, but there is to many reformatting of the code, so I can't merge it directly: https://github.com/CANopenNode/CANopenNode/pull/487

CANopenNode:
And please make it compliant with some [Misra C rules](https://github.com/CANopenNode/CANopenNode/blob/master/codingStyle), which was added recently: https://github.com/CANopenNode/CANopenNode/commit/a52db428f699394f58e12ce15acadfe1461e50ea

temi54c1l8:
ready

PR #505, opened 2024-03-14
Porting SRDO to v4.0

temi54c1l8:
Reference Issue: https://github.com/CANopenNode/CANopenNode/issues/503


CANopenNode:
Thank you. I need few days for review.

CANopenNode:
I'm late as I'm a little short with my time. Some dependencies will have to be removed:

- remove `#include "301/CO_SDOserver.h"`, `CO_SDOserver_t *SDO` from init functions,
- add `#include "301/CO_ODinterface.h"`, `OD_t *OD,` to init function

- remove `#include "301/CO_NMT_Heartbeat.h"`
- add `bool_t NMTisOperational` as argument to process function.

Please make this changes, if you can, I will help if necessary. Otherwise I can do this next week.

----
`#if (CO_CONFIG_SRDO) & CO_CONFIG_TSRDO_CALLS_EXTENSION` ....
this part may be more tricky, as it is totally new in v4 of CANopenNode. But there should be no dependency to SDO server.

CANopenNode:
Those `OD_entry_t *` arguments are not handled properly, etc. This OD interface is new in v4, it is powerful but may be complicated if not fully understood. Besides that SRDO is almost as complicated as PDO, which is the most.

The idea is the following: `CO_SRDO_init` accepts two arguments with `OD_entry_t *`. It should only get values from there and then set internal variables and configure the SRDO correctly. Internal variables are not related to `OD_entry_t *` any more then.

Also copying data should be similar as in PDO, with or without `CO_CONFIG_PDO_OD_IO_ACCESS` option, see: https://github.com/CANopenNode/CANopenNode/blob/master/301/CO_PDO.c#L827-L881

I would like for SRDO to be fully integrated before merging it. I would also like for SRDO to bi finally ported, but I'm quite short with my time. Please let me know, if you like me to make that part, hopefully in next ten days, then you test it?

temi54c1l8:
Great! As soon as you've done CALLS_EXTENSION and OD_IO_ACCESS I'll do all the tests!

I have already modified the dependencies as you indicated.

CANopenNode:
I merged the commits into main. It is not finished yet. There is no other changes, except SRDO.

CANopenNode:
Just for info: Currently I'm working quite intensively on it.

CANopenNode:
It is all done. It is quite a lot rewritten. Compiles well, but no tests were made at this point, probably won't work from the start.

temi54c1l8:
wow that's great, I'll start testing it right away

temi54c1l8:
see https://github.com/CANopenNode/CANopenNode/pull/516

temi54c1l8:
see https://github.com/CANopenNode/CANopenNode/pull/518

temi54c1l8:
see https://github.com/CANopenNode/CANopenNode/pull/520

PR #517, opened 2024-06-03
Fix srdo second part

temi54c1l8:
These are the second changes to make SRDO work correctly in Rx direction


CANopenNode:
Thank you very much. I just run out of time last week.

I added one more commit for fixing another cast data after bitwise.

PR #520, opened 2024-06-06
Only if the configuration is valid, the private variable is activated

temi54c1l8:
Only if the configuration is valid, the private variable is activated
(otherwise CO_SRDO_process will generate errors)


temi54c1l8:
CO_SRDO_init does not call configurationValidUnset if the configuration is invalid.

I preferred to write the control code in CO_SRDO_init_start because it is executed not only once for SRDOGuard (and not for all SRDOs)...

CANopenNode:
You are right, I totally missed that.

PR #521, opened 2024-06-06
Em

smartel99:


CANopenNode:
I don't agree with removal of `#if defined OD_DEFINITION` from `301/CO_ODinterface.h`

PR #522, opened 2024-06-10
Static analysis

temi54c1l8:
I started the static analysis with MISRA 2012 rules, in the next days I will make some commits


temi54c1l8:
In the last commit https://github.com/CANopenNode/CANopenNode/pull/522/commits/0142f418576c24d0006bc8bdcc9e59c2c25b6623  you could change the enums with define macros, sure it's not nice, but MISRA does not permit the use of bitwise operators '|' and '&' on enums, what do you think?

temi54c1l8:
If static analysis is too intrusive, consider cherry picking SRDO commits

CANopenNode:
This will raise the quality of the code. It doesn't matter, if there are many changes. I will take time and carefully verify each commit.

temi54c1l8:
@CANopenNode I HAVE COMPLETED THE STATIC ANALYSIS

I have excluded the OD.c and OD.h files from the check because I interpret them as configuration parameters (not source code execution)

I need help with the last 2 problems:

- file: CO_SDOserver.c ,line 640: bufShifted[countRd] = 0; /* (SDO->buf is one byte larger) */
  - report = CO_SDOserver: potential out of bounds pointer access: excess of 0:4294967269 bytes [MISRA 2012 Rule 18.1, required]
  - to solve this problem I usually put a: if(variable<XXX)

- file: CO_ODinterface.c ,line 119: (void)memcpy((void *)dataOrig, (const void *)buf, dataLenToCopy);
  - report = possible access beyond array for function 'memcpy(void *__restrict, const void *__restrict, size_t)', argument 3 (size=0:4294967295) exceeds argument 2 (size=4); during specific walk CO_PDO.c:627 OD_writeOriginal
  - this is not required by MISRA, I don't understand the report indication, if you can check

CANopenNode:
> file: CO_SDOserver.c ,line 640: bufShifted[countRd] = 0; /* (SDO->buf is one byte larger) */
> 
> * report = CO_SDOserver: potential out of bounds pointer access: excess of 0:4294967269 bytes [MISRA 2012 Rule 18.1, required]
> 
> * to solve this problem I usually put a: if(variable<XXX)

Maybe this: 
```
        if (countRd > 0 && countRd < ((CO_CONFIG_SDO_SRV_BUFFER_SIZE + 1) - countRemain) &&  (SDO->OD_IO.stream.attribute & ODA_STR) != 0) {
            bufShifted[countRd] = 0; /* (SDO->buf is one byte larger) */
            ...
```

> file: CO_ODinterface.c ,line 119: (void)memcpy((void *)dataOrig, (const void *)buf, dataLenToCopy);
> 
> * report = possible access beyond array for function 'memcpy(void *__restrict, const void *__restrict, size_t)', argument 3 (size=0:4294967295) exceeds argument 2 (size=4); during specific walk CO_PDO.c:627 OD_writeOriginal
> 
> * this is not required by MISRA, I don't understand the report indication, if you can check

Maybe additional check is expected: 
```
        if ((dataLenToCopy + stream->dataOffset) <= stream->dataLength && dataLenToCopy <= count) {
            (void)memcpy((void *)dataOrig, (const void *)buf, dataLenToCopy);
        }
        else {
            return ODR_DEV_INCOMPAT;
        }
```

CANopenNode:
Please add a description about MISRA rules applied to the stack. If there is some configuration file, it can be added to the base directory. Or file named "misra.md" could be added to the "doc" directory, describing something like: information about the tool used, misra version or something; which rules are used; some report summary; date of the last testing, ...

----

I have a question about number of commits: maybe it is hard to manage such large number of commits, I would prefer smaller number of commits or even single commit for all. It is possible to rebase, squash them. Github offers squashing 120 commit into single or something can be done manually. Please tell your opinion.

----

At this point, where may lines were already changed, we can also reformat the code wit auto formater. There is already a good `.clang-format` file from https://github.com/MaJerle/c-code-style. This would be another PR, but tell your opinion.

temi54c1l8:
For static analysis on a single commit I pulled it here: https://github.com/CANopenNode/CANopenNode/pull/525

I created the MISRA.md file and made and implemented the fixes discussed

Note: I am used to doing many commits (as it should be done with GIT) to explain the reason for the change, but I understand that for static analysis it is not necessary

temi54c1l8:
I don't know the .clang-format, I've always used https://github.com/uncrustify/uncrustify.
However, it's a good idea to reformat the code
in the next few days I'll try to see the .clang-format

martinwag:
please keep the individual commits. this will make it easier to find the actual reason for a change instead of just "MISRA".

martinwag:
In your md, you mention that you use vector pclintplus. As this is a commercial tool, most of the contributors can't check for compliance before committing. Did you have a look at the open source alternatives?

temi54c1l8:
Yes, it is correct, for SRDO (safety relevant data object) a certified tool (SIL qualified) is required.
I have not found alternative open source certified tools

CANopenNode:
If you two prefer individual commits, then I will merge that.
I don't see problems, if code is checked with commercial tools. Only I would add a line with a date of the last testing into the misra.md. So Misra compliance is guaranteed for commit at that date. This is good enough for the start.

temi54c1l8:
Individual commits allow for better change tracking.
I agree with adding a line with a last test date in misra.md.

CANopenNode:
I'm going through changes. Commit https://github.com/CANopenNode/CANopenNode/pull/522/commits/29bb7992d2c25d1deda78f4d7cb15fcbcf08e219 adds some unwanted behavior. Moving `#if` directive inside `case` won't respond with abort message, if segmented transfer is not supported.

To satisfy the Misra rule, this can be added before `default:', for example
```
#if ((CO_CONFIG_SDO_SRV) & CO_CONFIG_SDO_SRV_SEGMENTED) == 0
case CO_SDO_ST_DOWNLOAD_SEGMENT_REQ:
etc.
#endif
same for CO_CONFIG_SDO_SRV_BLOCK...
```

Or enum `CO_SDO_state_t` could be changed to `#define` statements.

That rule was not used in CO_SDOclient.c?

temi54c1l8:
You're right, I'll fix the code now

For SDOclient it's my problem because after the first changes I changed the driver header for a build, if you can wait a few days I'll complete the static analysis on everything

301/CO_fifo.h
```diff
@@ -471,64 +471,61 @@ size_t CO_fifo_readB642a(CO_fifo_t *fifo, char *buf, size_t count, bool_t end);
...
-    /** Bitmask for error bits */
-    CO_fifo_st_errMask = 0xF0U
-} CO_fifo_st;
+
+
+/**
+ * @defgroup uint8_t Bitfields for status argument from CO_fifo_cpyTok2U8 function and similar
+ * @{
+  */
+#define CO_fifo_st_closed   0x01U  /** Bit is set, if command delimiter is reached in src */
```
CANopenNode, line 480: According to doxygen, documentation comments after a macro or variable must begin with `/**<`
for example:
```
#define CO_fifo_st_closed   0x01U  /**< Bit is set, if command delimiter is reached in src */
```

temi54c1l8: fix on commit https://github.com/CANopenNode/CANopenNode/pull/522/commits/11771f889c75c6f9d78e69d1fc0ffe22cf451604

CANopenNode:
I will wait few days, until you finish. Thank you for your effort so far.

It seems that enumerators are not very practical with Misra c. I think, your solution is good. Maybe macros may be better also for `CO_SDO_state_t` enumerator.

You don't need to hurry with reformatting. After finishing with Misra I will try to auto-format the code and make a PR here for preview. I will also check doxygen documentation.

CANopenNode:
I managed to add three commits into your branch. Please don't forget to pull them before you make a next commit.

In PDO.c I reverted the behavior. If event timer or similar is missing in the object dictionary, CANopen won't break.

temi54c1l8:
@CANopenNode check this commit if it is correct https://github.com/CANopenNode/CANopenNode/pull/522/commits/b39334d9c3c3c3bd628eee48abc93621d5644014

CANopenNode:
> @CANopenNode check this commit if it is correct [b39334d](https://github.com/CANopenNode/CANopenNode/commit/b39334d9c3c3c3bd628eee48abc93621d5644014)

You mean additional checks against CO_CONFIG_SDO_CLI_BUFFER_SIZE? It seems correct.

CANopenNode:
> @CANopenNode 'sprintf' and 'snprintf' are deprecated. [MISRA 2012 Rule 21.6, required]
> I propose to make a macro, as suggested here: https://forum.misra.org.uk/archive/index.php?thread-1412.html
> If it's okay with you I'll proceed with the changes, or I'll add misra exclusions (maybe better)

You can exclude `#ifdef CO_DEBUG_SDO_CLIENT` and `#ifdef CO_DEBUG_SDO_SERVER` from static analysis. For other usages I recommend misra exclusion (fifo and gateway). Later we may replace s(n)printf with something else and remove exclusion.

temi54c1l8:
@CANopenNode Ho completato l'analisi statica cone tutte le configurazioni, l'unico errore che mi ho è il seguente:
https://github.com/CANopenNode/CANopenNode/pull/522#issuecomment-2194523048

Attenzione a questa raccomandazione: https://github.com/CANopenNode/CANopenNode/pull/522/commits/4a59d6eb0e35e62d258ab66ecac9ec6808afd88b

I will start again on Monday (I have a problem on SRDO with Codesys)

CANopenNode:
> @CANopenNode I need help with the last problem:
> File: CO_storageEeprom.c - line 148: uint32_t signatures[entriesCount];
> Variable length array 'signatures' of type 'uint32_t [entriesCount]' declared [MISRA 2012 Rule 18.8, required]

It is fixed with my last two commits.

temi54c1l8:
> > @CANopenNode I need help with the last problem:
> > File: CO_storageEeprom.c - line 148: uint32_t signatures[entriesCount];
> > Variable length array 'signatures' of type 'uint32_t [entriesCount]' declared [MISRA 2012 Rule 18.8, required]
> 
> It is fixed with my last two commits.

Great, I added the 'U' on the macro created for static analysis.
For me the static analysis is finished

temi54c1l8:
@CANopenNode 
After static analysis, I would like to pull these commits:
- https://github.com/temi54c1l8/CANopenNode/commit/240dd588fc0eef9d7a551b0e4d7c854f959d4ced
- https://github.com/temi54c1l8/CANopenNode/commit/28d3c1f372bb2275115b1b00da4d0322d27a86bb
- 
I ask you if you can check the implementation...

CANopenNode:
You made CO_SRDO_config() from part of CO_SRDO_init(). CO_SRDO_config() is called on transition to NMT operational.

I think, I made to optimized code and your approach is correct. According to standard SRDO should be configured on transition to operational... (I thought, original CO_SRDO_init() could be used on transition to NMT operational too, but your approach is better.)

Please make a PR, I will verify and merge this PR first.

temi54c1l8:
The commits do not start from the master branch but from the static_analysis branch, I merged them on this PR.
If you want them on the master I create new commits, but something will conflict and you have to solve them manually

CANopenNode (approved):

CANopenNode:
It is all merged into master, thank you for this large change.

temi54c1l8:
> It is all merged into master, thank you for this large change.

Thank you for this beautiful project

PR #548, opened 2024-10-17
Remove unreachable statement in CO_ODinterface.c

vinchatl:
The removed `break;` statement is unreachable and currently generates an error when compiled with ARM compiler (warning 111-D).

### DIFF
--- a/301/CO_ODinterface.c
+++ b/301/CO_ODinterface.c
@@ -182,13 +182,15 @@ OD_getSub(const OD_entry_t* entry, uint8_t subIndex, OD_IO_t* io, bool_t odOrig)
         return ODR_DEV_INCOMPAT;
     }
 
+    ODR_t ret = ODR_OK;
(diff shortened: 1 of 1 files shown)


CANopenNode:
This solution would break some Misra rules.
If we want to satisfy both rules, then function may be changed so, it returns only at the end. I suggest:

1. put `ODR_t ret = ODR_OK`
2. replace returns with `ret = ...`
3. use multiple `if (ret == ODR_OK) {` in parts of the function.
4. at the end `return ret;`

vinchatl:
I've never used the Misra rules but I afraid it could break one of these.

Your suggestion sounds sensible so I'll try to implement it. Do you prefer me to remove the 1st commit with a hard reset or to revert it in subsequent commits ?

CANopenNode:
Thank you. Guthub has an option to "squash and merge", so it can make single commit automatically.

PR #562, opened 2025-01-16
minor indentation code

temi54c1l8:


CANopenNode:
If applying this, auto-formatter will change it back. It is a little annoying anyway, because there is preprocessor directive between `} else` and `if (...) {`. I would leave unchanged.

temi54c1l8:
OK

PR #567, opened 2025-02-05
Better handling of SDOserver when multiple bytes needs to be written

DaMutz:
As far as I understand the custom 'write' command, it should return the number of bytes written, so the upper layer shall take this into account and executes the write several times, until the number of bytes have been written.

### DIFF
--- a/301/CO_SDOserver.c
+++ b/301/CO_SDOserver.c
@@ -465,13 +465,16 @@ validateAndWriteToOD(CO_SDOserver_t* SDO, CO_SDO_abortCode_t* abortCode, uint8_t
     /* may be unused */
     (void)crcOperation;
     (void)crcClient;
-    (void)bufOffsetWrOrig;
 
     /* write data */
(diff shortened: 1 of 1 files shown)


CANopenNode:
Actually this is bad solution, because it would block the operating system. CO_SDOserver_process() inside CO_process() is called inside loop (by OS), which completes the communication.

DaMutz:
I do not see how this is implemented, when the write function did not wrote everything. Does it make sense to lock and unlock on each loop iteration? If you think the path taken is wrong, I can also close the PR.

CANopenNode:
If write function did not wrote everything, then it usually have a reason - some internal buffer is full. All stack functions are non blocking, write is simply called again in next pass. do{}while will just block everything for undefined amount of time.

See https://github.com/CANopenNode/CANopenDemo/blob/master/tutorial/SDO.md#demo-domain for example of very large data transfer.

PR #569, opened 2025-03-07
Added support for CiA 401 / Driver example / Test example.

pedrorovi:
User must implement the CANInterface for its own hardware


martinwag:
please remove the vscode stuff.

CANopenNode:
We try to keep this project clean, without dependencies to hardware, extra examples, c++ dependencies, etc.

However, example with CiA401 would be welcome, maybe even better, if it is ready to work with some hardware.

You can just make your working project, use CANopenNode as a git submodule and publish it on your github profile with your own style, c++, vscode, etc.

I think, we could add a chapter into CANopenNode/README.md, named "Example projects" or something. Then we add links to interesting public projects there. We could also add some files to support CiA401 into CANopenNode, but with care.

PR #571, opened 2025-04-03
Extended CAN Ids

rgruening:
Hi, this pull request adds extended CAN Ids to the stack. The current design of the pull request makes them completely optional and not on by default.  
Extended Ids are not required by CANopen, but are supported in many COB-Ids. They are particularly useful when interfacing with non-CANopen protocols such as J1939.
The core change is a typedef, `typedef uint16_t CO_CANident_t`, that allows to widen (`uint32_t`) the CAN Rx/Tx buffer init functions for 29 bit identifiers.
~~I also introduced a new config flag `CO_CONFIG_FLAG_ALLOW_EXT_ID` which allows to disable the new behavior on a per module basis, not sure if this is of enough value or if it would be better to make the flag global for the whole stack.~~
If `CO_CONFIG_FLAG_ALLOW_EXT_ID` is not set, the stack should behave exactly as before the change.

Would love to hear your guys thoughts/feedback on the feature!


CANopenNode:
I took a quick look. It is a large PR and it seems professional and nice to me. It is a good idea to enable extended IDs.

It may take some time for me to review.

martinwag:
I also think this is a good idea. I had a look at all occurances of "uint16_t", as they might hint to a spot where the new type is missed. I found the following lines which in my opinion are suspicious:

`grep -rni . -e 'uint16_t'`

```
./CANopen.c:1201:            uint16_t preDefinedCanId = 0;;
./CANopen.c:1204:                preDefinedCanId = (uint16_t)((CO_CAN_ID_RPDO_1 + (i * 0x100U)) + nodeId);

./CANopen.c:1231:            uint16_t preDefinedCanId = 0;;
./CANopen.c:1234:                preDefinedCanId = (uint16_t)((CO_CAN_ID_TPDO_1 + (i * 0x100U)) + nodeId);


./304/CO_SRDO.c:168:        uint16_t defaultCOB_ID = SRDO->defaultCOB_ID + ((uint16_t)(stream->subIndex) - 5U);
./304/CO_SRDO.c:228:        uint16_t defaultCOB_ID = SRDO->defaultCOB_ID + index;
./304/CO_SRDO.c:662:             uint16_t defaultCOB_ID, OD_entry_t* OD_130x_SRDOCommPar, OD_entry_t* OD_138x_SRDOMapPar,

./304/CO_SRDO.h:137:    uint16_t defaultCOB_ID;          /**< From CO_SRDO_init() */
./304/CO_SRDO.h:218:                              uint8_t nodeId, uint16_t defaultCOB_ID, OD_entry_t* OD_130x_SRDOCommPar,

./301/CO_HBconsumer.c:184:        uint16_t COB_ID;
./301/CO_HBconsumer.c:198:            COB_ID = monitoredNode->nodeId + (uint16_t)CO_CAN_ID_HEARTBEAT;

./301/CO_SDOserver.c:295:    uint16_t CanId_ClientToServer, CanId_ServerToClient;

./301/CO_driver.h:225:    uint16_t ident; /**< Standard CAN Identifier (bits 0..10) + RTR (bit 11) */
./301/CO_driver.h:226:    uint16_t mask;  /**< Standard CAN Identifier mask with the same alignment as ident */

```
...and the places where those variables are used. The one in the example driver stands for discussion if the new type should be used there. @rgruening I think the other ones should use the new CO_CANident_t type, can you please have a look at those?

rgruening:
Btw, I noticed that in `CO_RPDO_init` (and TPDO) the COB-IDs are not checked if they are valid in the SDO Write functions for the COB-IDs they are (`OD_write_14xx`) is there an explicit reason? If you configure invalid Ids, even before this pull-request and you have `CO_CONFIG_FLAG_OD_DYNAMIC` disabled. You don't get an error but receive on complete different Ids.

rgruening:
Hi, thanks for testing, and good catch! You're right; the macro should only check standard IDs. Extended IDs in the standard range should also work now since they have the extended bit (0x20000000U) set, making them larger than 0x7FFU.

ant0nisk:
Perfect, this all works nicely. I can configure PDOs with the Extended IDs. And if I don't set bit-29, it's rejecting everything above 7FF.

Nice!

image

Issue #531, opened 2024-08-10
[feature] PDO bitwise mapping


Sl-Alex-KION:
Hello and thank you very much for this awesome project. I was using it in many projects, running on different microcontrollers and even on Android.
More than a half of the projects I'm working on use bitwise PDO mapping. I know that it is very easy to map the entire PDO to a single UINT64 object or to a union with a packed bit field, but this is not very convenient. Instead of having all the data defined in the object dictionary, we define them in the code.
I would like to volunteer and implement this feature. I see you invest o lot of time trying to reduce the amount of resources used by CANopenNode, so I'd do it in the following way:

1. Define ```CO_CONFIG_PDO_BITWISE_MAPPING``` in ```CO_config.h```, next to ```CO_CONFIG_PDO_OD_IO_ACCESS```. We still have some unused bits for this.
2. If this feature enabled, I'll reuse OD_IO.dataOffset to store the number of bits mapped instead of a number of bytes.
3. I'll take care about the following functions (hopefully didn't forget something, still need some time to debug all possible configurations):
    - ```PDOconfigMap```
    - ```PDO_initMapping```, both with and without ```CO_CONFIG_PDO_OD_IO_ACCESS```
    - ```OD_write_PDO_mapping```
    - ```CO_PDO_receive```
    - ```CO_RPDO_process```
    - ```CO_TPDO_process```
4. I'll take care of big-endian platforms. I don't have any big-endian HW to test my changes, but this should not be a big deal because PDOs are quite simple.

I already have some working skeleton I'd like to polish, test and share with you in a couple of weeks, so I have a couple of small questions:
- Would you accept such a PR?
- Do you plan any significant changes in the project (structure, code style, docs) in the nearest future?
- Is the idea with the additional config option fine for you, or you have another option?

Please let me know, it would be a real pleasure for me to make the perfect project even better. Even some commercial CANopen stacks don't have this feature.

Sl-Alex:
After this change is implemented, I'll need to add it to the Object Dictionary Editor, I already have a working skeleton there, so this could be a step 2.

CANopenNode:
> * Would you accept such a PR?

Yes, as long as PR is according to the standard, it is welcome.

> * Do you plan any significant changes in the project (structure, code style, docs) in the nearest future?

Recently there was significant changes (misra, reformatting) in the code, but there are no more similar plans. 

> * Is the idea with the additional config option fine for you, or you have another option?

[...]

PR #572, opened 2025-04-06
Implement PDO bitwise mapping

Sl-Alex:
This PR implements my favorite [bitwise mapping feature for PDOs](https://github.com/CANopenNode/CANopenNode/issues/531).

It is a configuration option, so it can be disabled to increase the speed if needed.
Tested on big-endian (PPC e500 on QEMU) and little-endian (native Windows) platforms.

~~As of now, OD editor doesn't support bitwise mapping, so if you need this feature, you would need to manually adjust the mapping in ```OD.c```. Bitwise PDO mapping support in OD editor would be my next task.~~
UPD: Implemented, see this PR: https://github.com/CANopenNode/CANopenEditor/pull/198

### DIFF
--- a/301/CO_PDO.c
+++ b/301/CO_PDO.c
@@ -295,7 +295,7 @@ PDO_initMapping(CO_PDO_common_t* PDO, OD_t* OD, OD_entry_t* OD_PDOMapPar, bool_t
         }
         return CO_ERROR_OD_PARAMETERS;
     }
-    if (mappedObjectsCount > CO_PDO_MAX_SIZE) {
+    if (mappedObjectsCount > CO_PDO_MAX_MAPPED_ENTRIES) {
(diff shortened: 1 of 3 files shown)


CANopenNode:
I was short of time for CANopen, so this PR and CANopenEditor is waiting. I may now have some time and I have plans to put later in better shape. I think, currently CANopenEditor is too messy and is very hard to add "simple things"...

CANopenNode:
Thank you for both. Bitwise mapping is now merged into CANopenNode and into https://github.com/CANopenNode/CANopenEditor/pull/198

It is disabled by default, so there is no change in existing usage. To enable bitwise PDO mapping use CO_CONFIG_PDO_BITWISE_MAPPING configuration macro (see CO_config.h).

PR #590, opened 2025-07-25
EMCY message error on CO_EM_RPDO_WRONG_LENGTH

temi54c1l8:
The compliance test that requires the bit on the EMCY message error register to be activated as a result of the arrival of RPDO1 with wrong data lenght


CANopenNode:
Thank you for your PRs. There is quite some PRs waiting to be merged. I'm a currently a little lack of time. I plan to verify and merge them in the near future. I will indicate all those updates with a new version. It won't change stack much, but anyway.

PR #627, opened 2026-05-31
fix: gateway tok[20] too small — lss_activate_bitrate always fails

### LINKED ISSUE #623: lss_activate_bitrate command fails:  tok[20] buffer is  too small
"lss_activate_bitrate"  is 20 chars and the token buffer is only 20 bytes

tokenSize == count 
    *err = true;|

tbitcs:
## Problem

The local command token buffer in CO_GTWA_process() is char tok[20]. The command string lss_activate_bitrate is exactly 20 characters.

CO_fifo_readToken() is called with count = sizeof(tok) = 20. Inside that function, when 	okenSize == count the overflow guard fires: *err = true and the token is discarded as if it overflowed. The command therefore **always fails with a syntax error** and can never be executed.

Root cause (from 301/CO_fifo.c):
`c
if (tokenSize == count) {   // overflow detected
    tokenSize = 0;          // discard
    *err = true;
}
`

The buffer needs at least **21 bytes** (20 chars + null terminator).

## Fix

One-character change: char tok[20] → char tok[21].

Fixes #623.


CANopenNode:
You are correct, thank you.

